PR #2, opened 2019-04-25
Fs apr19 vadims

Kresh24107:


src/style.css
```diff
@@ -0,0 +1,84 @@
...
+  width: 100%;
+  padding-top: 12px;
+  padding-bottom: 12px;
+  margin-top: 20px;
+}
+
+button:hover {
+  color: #00acdc;
+  background: #fff;
+  box-sizing: border-box;
```
ivadimko: don't set such styles for `:hover` state. It should be set for the default state. Even better to set if for all items on the page to prevent unexpected behavior on other blocks

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+}
+
+.card {
+  width: 198px;
```
ivadimko: In description card's width should be 200px

ivadimko: figure out how to fix it and make the borders a part of the card's size

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+.code {
+  font-size: 11px;
+  line-height: 13px;/* identical to box height */
+  margin: 6px 0 15px;
+}
+
+.review > img {
+  float: left;
+  width: 13px;
+  height: 14px;
```
ivadimko: image is 15x15 in the mock. Don't use such unexpected styles

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+  padding: 0 5px 0 1px;
+}
+
+.review > p {
+  font-size: 11px;
+  line-height: 13px;/* identical to box height */
+  text-align: right;
+  padding-top: 1px;
+  margin-bottom: 25px;
```
ivadimko: why did you set margin-bottom for `<p>` instead of setting it to `.review`?

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+}
+
+.card {
+  width: 198px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+
+.product {
+  padding: 30px 20px;
```
ivadimko: It's unexpected behavior with `padding` on the image. Better to wrap it in separate block and set styles for the wrapper block.

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+  width: 158px;
+}
+
+.wrap {
+  padding: 11px 15px 15px;
```
ivadimko: What does the `padding-top: 11px` do in this case? Better to set `margin-bottom: 45px` for the image wrapper. As in the mock.
image

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+  font-size: 11px;
+  line-height: 13px;/* identical to box height */
+  margin: 6px 0 15px;
+}
+
+.review > img {
+  float: left;
+  width: 13px;
+  height: 14px;
+  padding: 0 5px 0 1px;
```
ivadimko: the distance between stars is `4px`. Try to figure it out with `margin-right: 4px` 
image

ivadimko: Make sure the star is centralized vertically and horizontally in its wrapper

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+.left {
+  font-size: 11px;
+  line-height: 15px;/* identical to box height */
+  float: left;
+}
+
+.right {
+  font-weight: bold;
+  font-size: 16px;
+  line-height: 11px;/* identical to box height */
```
ivadimko: box height is 19px
image

src/style.css
```diff
@@ -0,0 +1,84 @@
...
+  width: 13px;
+  height: 14px;
+  padding: 0 5px 0 1px;
+}
+
+.review > p {
+  font-size: 11px;
+  line-height: 13px;/* identical to box height */
+  text-align: right;
+  padding-top: 1px;
```
ivadimko: what is it? The text should be centralized vertically with stars. Figure out how to do it with `flexbox`

src/index.html
```diff
@@ -11,9 +11,29 @@
...
+      <img src="images/star-active.svg" alt="star-active">
+      <img src="images/star-active.svg" alt="star-active">
+      <img src="images/star.svg" alt="star">
+      <p>Отзывов: 5</p>
+    </div>
+    <div class="price">
+      <span class="left">Цена:</span>
+      <span class="right">69 999 грн</span>
+    </div>
+    <button data-qa="hover">Купить</button>
```
ivadimko: In the task description, there is a point that it should be a link. Please read the description carefully before starting.
image

src/index.html
```diff
@@ -11,9 +11,29 @@
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <title>Document</title>
   <link rel="stylesheet" href="./style.css">
+  <link href="https://fonts.googleapis.com/css?family=Roboto" rel="stylesheet">
```
ivadimko: Please define here a font-weights which do you want to use. Now you load only `font-weight: regular` (or `font-weight: 400`). But there are also `medium (500)` and `bold(700)` in the mock
image

Then the link will look like this.
image

ivadimko:
Overall good job. Looks great. Just need to add correct fonts.
Some minor changes requested.

src/index.html
```diff
@@ -11,9 +11,29 @@
...
 </head>
 <body>
-
+<div class="card" data-qa="card">
+  <img src="images/imac.jpeg" alt="imac" class="product">
+  <div class="wrap">
+    <p class="article">Моноблок APPLE A1419 iMac 2727" Retina 5K (MNED2UA/A)</p>
```
ivadimko: 27", not 2727"

src/index.html
ivadimko: By the way, `button` takes System UI font. So you have to overwrite it explicitly. Now your button has this font (the Apple ones)
image

ivadimko:
I don't see `clear` after `float`

ivadimko:
I've checked your layout, fixed fonts and it started working. I didn't change any margins / paddings etc
image

src/style.css
```diff
@@ -0,0 +1,84 @@
+/* No passed !!! */
+
+body {
+  margin: 0;
+  font-family: "Roboto", "sans-serif";
```
ivadimko: `sans-serif` should be without quotes. Otherwise, it fails to apply that font.

PR #3, opened 2019-04-26
Vitalii Pshychenko

VitaliiPshychenko:


src/style.css
```diff
@@ -0,0 +1,109 @@
...
+  width: 15px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.rating-text {
+  font-size: 11px;
+  line-height: 14px;
+  padding-top: 2px;
```
ivadimko: don't use such kind of `hacks`. The aim here is to align text vertically in the middle.

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+.price-title {
+  font-size: 11px;
+  line-height: 13px;
+}
+
+.cost {
+  font-weight: bold;
+  font-size: 16px;
+  line-height: 19px;
+  width: 79px;
```
ivadimko: Don't use such kind of explicitly defined width. The card should be flexible. One item can have price `10` and another - `10000000`. The width should not be set.

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+}
+
+.buy {
+  border: 1px solid #00acdc;
+  box-sizing: border-box;
```
ivadimko: move it to general styles. Further, you'll use some `reset.css` and all of such styles you will need to move there.
https://bitsofco.de/a-look-at-css-resets-in-2018/

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+  line-height: 16px;
+  text-align: center;
+  text-decoration: none;
+  text-transform: uppercase;
+  color: #fff;
+  background: #00acdc;
+  border-radius: 5px;
+  padding: 12px 0;
+  display: block;
+  transition: 0.5s ease;
```
ivadimko: Great job. Just a suggestion: define explicitly which property would you like to be animated. 
```css
transition: color 0.5s ease, background-color 0.5s ease;
```
or: 
```css
transition-property: color, background-color;
transition-duration: 0.5s;
transition-timing-function: ease;
```

https://developer.mozilla.org/en-US/docs/Web/CSS/transition

src/style.css
```diff
@@ -0,0 +1,107 @@
...
+  display: flex;
+  margin: 0;
+  
+}
+
+.rating {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 13px 0 19px;
```
ivadimko: Why these numbers? There are 15px 0 20px
image
image

src/style.css
```diff
@@ -0,0 +1,107 @@
...
+.product-card {
+  background: #fff;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  padding: 30px 15px 15px;
+  width: 200px;
+}
+
+.image-wrap {
+  height: 134px;
```
ivadimko: Since you set height explicitly to this block you have to center the image vertically and horizontally. Try to figure it out with `flexbox`

PR #30, opened 2019-05-03
Vladimir`s_Solution

VladimirTi:


src/index.html
```diff
@@ -9,11 +9,34 @@
           maximum-scale=1.0,
           minimum-scale=1.0">
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <link href="https://fonts.googleapis.com/css?family=Roboto:400,500,700" rel="stylesheet">
   <title>Document</title>
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
+  <div class="card" data-qa="card">
+    <img src="images/imac.jpeg" class="product" alt="">
```
gestap: you need to fill out `alt`

src/index.html
```diff
@@ -9,11 +9,34 @@
...
+      </div>
+        <span>Отзывов: 5</span>
+    </div>
+
+    <div class="margins flexbox between price">
+      <p class="price-text">Цена:</p>
+      <p class="value">69 999 грн</p>
+    </div>
+
+    <button>купить</button>
```
gestap: `Item with text 'Купить' should be a link.`

src/style.css
```diff
@@ -0,0 +1,94 @@
...
+  align-items: flex-end;
+}
+
+.between {
+  justify-content: space-between;
+}
+
+.margins {
+  margin-left: 15px;
+  margin-right: 15px;
```
gestap: Better to write it in one line 
`margin: 0 15px;`

src/index.html
```diff
@@ -9,11 +9,34 @@
...
+      </div>
+        <span>Отзывов: 5</span>
+    </div>
+
+    <div class="margins flexbox between price">
+      <p class="price-text">Цена:</p>
+      <p class="value">69 999 грн</p>
+    </div>
+
+    <a href="#"><button>купить</button></a>
```
gestap: I've to recommend using only `a` without `button`.

src/style.css
VladimirTi: this variant will not work due to conflict with margin-top and margin-bottom in other`s rules

src/index.html
```diff
@@ -9,11 +9,37 @@
           maximum-scale=1.0,
           minimum-scale=1.0">
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <link href="https://fonts.googleapis.com/css?family=Roboto:400,500,700" rel="stylesheet">
   <title>Document</title>
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
+  <body>
```
gestap: why you have two `body`?

PR #46, opened 2019-05-24
Fe mar19 olena

OlenaZubashenko:


src/index.html
```diff
@@ -1,19 +1,59 @@
...
+      <img src="./images/imac.jpeg" alt="imac">
+    </div>
+    <div class="wrapper__discription">
+      Моноблок APPLE A1419 iMac
+      27" Retina 5K (MNED2UA/A)
+    </div>
+    <div class="wrapper__code">Код товара: 195434
+    </div>
+    <div class="wrapper__rating">
+      <div class="wrapper__rating__group">
```
lipskiy-aleh: Why did you call this class like a group?  This class response only for one star in your code.

src/index.html
```diff
@@ -1,19 +1,59 @@
...
+        <img src="./images/star-active.svg" alt="">
```
lipskiy-aleh: This is bed idea to use tag img here. Because if you need change style for active to non-active so you need to manipulate with HTML, batter use css class which will be responsible for active and non-active styles.

src/style.css
```diff
@@ -0,0 +1,109 @@
+body {
+  font-family: "Roboto", sans-serif;
+  margin: 0;
+}
+
+.wrapper {
+  display: block;
```
lipskiy-aleh: Why do you need this property for div ?

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+  font-size: 12px;
+  line-height: 15px;
+  color: black;
+  padding-bottom: 5px;
+  padding-top: 41px;
+  margin: 0;
+}
+
+.wrapper__code,
+.wrapper__cost__text {
```
lipskiy-aleh: Batter to use separate set of rules for .wrapper__code and .wrapper__cost__text

src/index.html
```diff
@@ -1,19 +1,59 @@
...
+      <div class="wrapper__rating__views">
+        Отзывов: 5
+      </div>
+    </div>
+    <div class="wrapper__cost">
+      <div class="wrapper__cost__text">Цена:</div>
+      <div class="wrapper__cost__num">69 999 грн</div>
+    </div>
+    <div class="wrapper__cost__btn">
+      <a href="#" class="wrapper__cost__btn__link" data-qa="hover">Купить</a>
```
lipskiy-aleh: As for me batter to use the button for this action

PR #123, opened 2019-07-27
Product card done!

Vladyslav223:


src/index.html
```diff
@@ -11,9 +11,29 @@
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <title>Document</title>
   <link rel="stylesheet" href="./style.css">
+  <link href="https://fonts.googleapis.com/css?family=Roboto:400,500&display=swap" rel="stylesheet">
 </head>
 <body>
-
+  <div class="product-card" data-qa="card">
```
danheim: Use `section` tag

src/index.html
```diff
@@ -11,9 +11,29 @@
...
+  <img class="product-card__image" src="images/imac.jpeg" alt="imac" >
```
danheim: Wrap it in `<a>` tag

src/index.html
```diff
@@ -11,9 +11,29 @@
...
+  <h2 class="product-card__heading">Моноблок APPLE A1419 iMac 27" Retina 5K (MNED2UA/A)</h2>
+  <span class="product-card__code">Код товара: 195434</span>
+  <div class="product-card__reviews">
+    <div class="product-card__reviews__rate">
+      <img src="images/star-active.svg" alt="">
```
danheim: This is not a logic image, implement it only with css (background-image)

src/style.css
```diff
@@ -0,0 +1,93 @@
...
+  margin-bottom: 45px;
+}
+
+.product-card__heading {
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 15px;
+  margin-bottom: 5px;
+  margin-top: 0;
```
danheim: don't repeat margin property
```suggestion
  margin: 0 0 15px;
```

margin: top right bottom left

src/style.css
```diff
@@ -0,0 +1,93 @@
...
+  font-size: 11px;
+  background: #fff;
+  border: 1px solid #f3f3f3;
+  box-sizing: border-box;
+  border-radius: 5px;
+}
+
+.product-card__image {
+  width: 160px;
+  height: 134px;
```
danheim: To best sizing try to use only width or height.

src/index.html
Vladyslav223: In readme it was no word about it

src/style.css
Vladyslav223: I did not understand about margins. Here i want to set only top and bottom magins, and i do not want to set a left and right margins, how should i write?

src/style.css
Vladyslav223: Using one of this properties not possible to pass test. If i use both- it pass perfectly

Vladyslav223: I found out the problem. With width test not pass, only with height. Why i dont know..

src/style.css
irenhh: since your heading doesn't have default margins for left and right side, you can just you here shorthand property, you'll not break anything in this case. Shorthand property is better to use if possible

PR #125, opened 2019-07-28
Fs jul19 yevhen sag - product-cards solution

eugeneSahach:
- [DEMO LINK](https://yevhen-sahach.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://yevhen-sahach.github.io/layout_product-cards/report/html_report/)


src/index.html
```diff
@@ -11,9 +11,29 @@
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <title>Document</title>
   <link rel="stylesheet" href="./style.css">
+  <link href="https://fonts.googleapis.com/css?family=Roboto:400,500,700&display=swap" rel="stylesheet">
 </head>
 <body>
-
+  <div class="product-card" data-qa="card">
+    <img src="images/imac.jpeg" alt="imac" class="product-image">
+    <p class="product-name">Моноблок APPLE A1419 iMac 27" Retina 5K (MNED2UA/A)</p>
```
danheim: It should be a header

src/index.html
```diff
@@ -11,9 +11,29 @@
...
+    <p class="product-code">Код товара: 195434</p>
+    <div class="feedbacks-line">
+      <div class="average-rate">
+        <img src="images/star-active.svg" alt="active-star" class="star">
+        <img src="images/star-active.svg" alt="active-star" class="star">
+        <img src="images/star-active.svg" alt="active-star" class="star">
+        <img src="images/star-active.svg" alt="active-star" class="star">
+        <img src="images/star.svg" alt="empty-star" class="star">
```
danheim: It can called just `star`

src/index.html
```diff
@@ -11,9 +11,29 @@
...
+      </div>
+      <p class="feedbacks">Отзывов: 5</p>
+    </div>
+    <div class="price-line">
+      <p class="price-text">Цена:</p>
+      <p class="price">69 999 грн</p>
+    </div>
+    <a class="order-button" data-qa="hover">купить</a>
```
danheim: Tag `a` should has a `href` attribute.

src/style.css
```diff
@@ -0,0 +1,90 @@
+.product-card {
+  box-sizing: border-box;
+  display: flex;
+  flex-flow: column nowrap;
```
danheim: It may be just `flex-direction: column`

src/style.css
```diff
@@ -0,0 +1,90 @@
...
+  width: 200px;
+  height: 403px;
+  background: #fff;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  font-family: Roboto;
```
danheim: Which font family will be displaying if google fonts not responding?
Please, add alternative common one, like `Arial` or `sens-serif`.

src/style.css
```diff
@@ -0,0 +1,90 @@
...
+  font-weight: normal;
+  font-size: 11px;
+  line-height: 13px;
+  padding: 14px 14px 15px;
+}
+
+.product-image {
+  width: 160px;
+  height: 134px;
```
danheim: To better sizing, try to use only width or height.

src/style.css
```diff
@@ -0,0 +1,90 @@
...
+  margin: 16px auto 0;
+}
+
+.product-name {
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 15px;
+  margin: 45px auto 0;
```
danheim: This is a flex element, try to centered using justify-content in a parent.

src/style.css
```diff
@@ -0,0 +1,90 @@
...
+}
+
+.product-code {
+  align-self: flex-start;
```
danheim: default property

src/style.css
yevhen-sahach: 160/134 px isn't original size of image, its size that I need for this card, if I remove width property it gets bigger

src/style.css
yevhen-sahach: I've tried, then it moves left for some reason

yevhen-sahach:
applied all sugestions excepr of two that I commented above

src/style.css
danheim: I meant you don't need use `auto` in `margin`
```suggestion
  margin: 45px 0 0;
```

src/style.css
```diff
@@ -0,0 +1,90 @@
...
+  height: 40px;
+  font-weight: bold;
+  font-size: 14px;
+  line-height: 16px;
+  color: #fff;
+  text-transform: uppercase;
+  text-decoration: none;
+  background: #00acdc;
+  border-radius: 5px;
+  margin-top: 20px;
```
danheim: `cursor: pointer;`

yevhen-sahach: it's default value for link, am I wrong?

PR #128, opened 2019-07-28
HW-9 solution

kosbielozorov:


src/style.css
```diff
@@ -0,0 +1,119 @@
...
+}
+
+.product-card:hover {
+  box-shadow: 0 2px 5px rgba(0, 0, 0, 0.07);
+}
+
+.product-pic {
+  display: block;
+  width: 160px;
+  height: 134px;
```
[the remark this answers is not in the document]
KosBelozyorov: i understand why you say me so, but if i use only height or width i have same errors in tests

KosBelozyorov:
fixed index.html and style.css

src/style.css
```diff
@@ -0,0 +1,119 @@
...
+  display: inline-block;
+  width: 100%;
+  height: 38px;
+  background-color: #00acdc;
+  border: 1px solid #00acdc;
+  border-radius: 5px;
+  color: #fff;
+  font-size: 14px;
+  font-weight: 700;
+  line-height: 38px;
```
danheim: I'm not sure you need it here.

KosBelozyorov: line-height: 38px; - this is a vertical align for button's title

PR #129, opened 2019-07-28
My solution to "Product card"

VladyslavSak:


src/style.css
```diff
@@ -0,0 +1,111 @@
...
+  display: block;
+  margin-top: 5px;
+}
+
+.product_reviews {
+  display: flex;
+  justify-content: space-between;
+  align-content: space-between;
+  margin-top: 15px;
+  margin-bottom: 20px;
```
danheim: Don't repeat margin property
```suggestion
  margin: 15px 0 20px;
```

src/style.css
```diff
@@ -0,0 +1,111 @@
...
+}
+
+.product_reviews-stars {
+  width: 91px;
+  display: flex;
+  justify-content: space-between;
+}
+
+.star {
+  background: url("./images/star-active.svg") center center no-repeat;
```
danheim: Don't repeat center

src/style.css
```diff
@@ -0,0 +1,111 @@
+body {
+  margin: 0;
+  background-color: #ccc;
+}
+
+h2 {
```
danheim: Style classes, not tags.

PR #169, opened 2019-09-03
Fs aug19 kirill

anisimovv:
Test don't work. All sizes and margins same as in Figma.


src/style.css
```diff
@@ -0,0 +1,98 @@
...
+}
+
+.product_card__price_val {
+  font-weight: bold;
+  font-size: 16px;
+}
+
+.product_card__button {
+  display: block;
+  width: 100%;
```
gestap: it's a default value for block

PR #171, opened 2019-09-04
solution

karpyshynroman:


src/style.css
```diff
@@ -0,0 +1,80 @@
+* {
+  box-sizing: border-box;
+  padding: 0;
+  margin: 0;
+  font-family: "Roboto", sans-serif, "Times New Roman";
```
gestap: don't add FF to `*`, add it to body

gestap:
Good job!

src/style.css
```diff
@@ -0,0 +1,83 @@
+* {
+  box-sizing: border-box;
+  padding: 0;
+  margin: 0;
+  text-decoration: none;
```
gestap: better to write like this

```
a {
  text-decoration: none;
}
```

gestap: no need to add this rule to all elements

PR #187, opened 2019-09-06
Fs aug19 sasha g

Shurkodr:


src/style.css
```diff
@@ -0,0 +1,120 @@
+body {
+  margin: 0;
```
gestap:
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,120 @@
...
+}
+
+.conten-box {
+  padding: 0 14px 15px;
+  display: flex;
+  flex-direction: column;
+}
+
+.product-name {
+  font-weight: 500;
```
gestap: you set this value by default in body, no need to write it again
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,120 @@
...
+.otzivi-box {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin-bottom: 20px;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
```
gestap: it's default value for flex
```suggestion
```

PR #314, opened 2020-03-09
solution

coastin:
- [LINK](https://coastin.github.io/layout_product-cards/)
- [REPORT LINK](https://coastin.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -10,10 +10,38 @@
           minimum-scale=1.0">
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <title>Document</title>
-  <link rel="stylesheet" href="./style.css">
+  <link href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700" rel="stylesheet">
+  <link rel="stylesheet" href="style.css">
 </head>
 <body>
-
-<script type="text/javascript" src="./main.js"></script>
+  <div class="container" data-qa="card">
+    <div class="imac">
+      <img src="images/imac.jpeg" alt="imac">
+    </div>
+    <div class="description">
+      <p>Моноблок APPLE A1419 iMac 27" Retina 5K (MNED2UA/A)</p>
+      <p id="itemCode">Код товара: 195434</p>
+    </div>
+    <div class="stars">
+      <div>
+        <img src="images/star-active.svg" alt="active star">
+        <img src="images/star-active.svg" alt="active star">
+        <img src="images/star-active.svg" alt="active star">
+        <img src="images/star-active.svg" alt="active star">
+        <img src="images/star.svg" alt="star">
+      </div>
+      <div id="feedback">
+        <span>Отзывов: 5</span>
+      </div>
+    </div>
+    <div class="price">
+      <span>Цена:</span>
+      <span>69 999 грн</span>
+    </div>
+    <button class="buy">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -10,10 +10,38 @@
...
+      <p>Моноблок APPLE A1419 iMac 27" Retina 5K (MNED2UA/A)</p>
+      <p id="itemCode">Код товара: 195434</p>
+    </div>
+    <div class="stars">
+      <div>
+        <img src="images/star-active.svg" alt="active star">
+        <img src="images/star-active.svg" alt="active star">
+        <img src="images/star-active.svg" alt="active star">
+        <img src="images/star-active.svg" alt="active star">
+        <img src="images/star.svg" alt="star">
```
vkryvytskyy: this should be done the same way, as in the `stars` task

src/index.html
```diff
@@ -10,10 +10,38 @@
...
 </head>
 <body>
-
-<script type="text/javascript" src="./main.js"></script>
+  <div class="container" data-qa="card">
+    <div class="imac">
+      <img src="images/imac.jpeg" alt="imac">
+    </div>
+    <div class="description">
+      <p>Моноблок APPLE A1419 iMac 27" Retina 5K (MNED2UA/A)</p>
```
vkryvytskyy: it should be a heading tag

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  flex-wrap: wrap;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+
+.imac {
+  height: 194px;
+}
+
+.imac img {
```
vkryvytskyy: Try to avoid stylizing tags. Always use classes for this. Tags stylizing is OK when you reset default browser styles, like margin: 0 for BODY tag, which has 8px margins.

vkryvytskyy: Also try to avoid increasing the specificity of the selector. One class is enough to identify the elements on the page. No need to write something like .block.colored-block. Use .block and .block--colored instead, and apply each of them to your element

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+
+.description p {
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 15px;
+  margin: 15px 0;
+  margin-bottom: 0;
+}
+
+#itemCode {
```
vkryvytskyy: IDs are used for JS, try not to style the id if you can avoid it. This makes impossible to reuse the code. Stylize classes. [Style guide](https://mate-academy.github.io/style-guides/htmlcss.html#id-selectors)

src/index.html
vkryvytskyy: heading tag

src/index.html
vkryvytskyy: fix this

src/style.css
vkryvytskyy: fix this

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  justify-content: space-between;
+  width: 100%;
+}
+
+.stars {
+  margin: 0 15px;
+  margin-top: 15px;
+}
+
+#feedback span {
```
vkryvytskyy: don't use ids for styling

PR #315, opened 2020-03-09
solution

olegobiukh:
[DEMO LINK](https://olegobiukh.github.io/layout_product-cards/)
 [TEST REPORT LINK](https://olegobiukh.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -5,9 +5,33 @@
   <meta name="viewport"
         content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
   <title>Product cards</title>
+  <link href="https://fonts.googleapis.com/css?family=Roboto:400,500,900&display=swap" rel="stylesheet">
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
-  <h1>Product cards</h1>
+  <div class="card" data-qa="card">
+    <div class="card__image-wrapper">
+      <img src="./images/imac.jpeg" alt="imac" class="card__image">
+    </div>
+    <div class="card__content">
+      <p class="card__title">Моноблок APPLE A1419 iMac 27" Retina 5K (MNED2UA/A)</p>
+      <span class="card__text">Код товара: 195434</span>
+      <div class="card__box ratings">
+        <div class="ratings__stars">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -5,9 +5,33 @@
...
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
-  <h1>Product cards</h1>
+  <div class="card" data-qa="card">
+    <div class="card__image-wrapper">
+      <img src="./images/imac.jpeg" alt="imac" class="card__image">
+    </div>
+    <div class="card__content">
+      <p class="card__title">Моноблок APPLE A1419 iMac 27" Retina 5K (MNED2UA/A)</p>
```
vkryvytskyy: there should be some of the heading tags

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  display: flex;
+  padding: 0 5px;
+  justify-content: center;
+  align-items: center;
+  /* margin-bottom: 45px; */
+}
+
+.card__image {
+  width: 100%;
+  /* width: 160px; */
```
vkryvytskyy: remove comments

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+
+.card__image-wrapper {
+  display: flex;
+  padding: 0 5px;
+  justify-content: center;
+  align-items: center;
+  /* margin-bottom: 45px; */
```
vkryvytskyy: remove comments

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+.ratings__star {
+  width: 15px;
+  height: 15px;
+  margin-right: 4px;
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.ratings__stars .ratings__star:nth-child(-n+4) {
```
vkryvytskyy: Try to avoid increasing the specificity of the selector. One class is enough to identify the elements on the page. No need to write something like .block.colored-block. Use .block and .block--colored instead, and apply each of them to your element

PR #363, opened 2020-07-06
Solution

rmnkk:
[DEMO LINK](https://rmnkk.github.io/layout_product-cards/) <br>
[TEST REPORT LINK](https://rmnkk.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,111 @@
+@import url("https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,300;0,500;0,700;1,400&display=swap");
+
+html {
+  font-family: "Roboto", "Arial", sans-serif;
+}
+
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.card {
+  width: 200px;
+  border: 1px solid #f3f3f3;
+  box-sizing: border-box;
+  padding: 30px 14px 14px;
+  border-radius: 5px;
+  display: flex;
+  flex-direction: column;
+}
+
+.card_image {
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,108 @@
...
+.stars__item {
+  width: 15px;
+  height: 15px;
+  margin-right: 4px;
+  background-image: url("./images/star.svg");
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.stars--4 .stars__item:nth-child(-n+4) {
```
vsm2k17, line 59: add classes from stars block task, so if I change the class of your element in dev tools I will see how the count of active stars changing.

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  background: #00acdc;
+  color: #fff;
+  border-radius: 5px;
+  box-sizing: border-box;
+}
+
+.card__button:hover {
+  color: #00acdc;
+  background: #fff;
+  border: 1px solid #00acdc;
```
vsm2k17, line 107: at this moment when you hovered on your link - border appears and adds 1px, because of box-sizing: border-box; width and height of your button is recalculating, and text could change his position, so its better if you write border property inside the button.

src/style.css
rmnkk: Okay. Fixed.

src/style.css
rmnkk: Got it. Fixed.

PR #372, opened 2020-07-07
Solution

krajnova:
- [DEMO LINK](https://krajnova.github.io/layout_product-cards/) <br>
- [TEST REPORT LINK](https://krajnova.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,111 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  font-family: Roboto, sans-serif;
+}
+
+.card {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  width: 200px;
+  padding: 30px 14px 14px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+
+.card__photo {
+  width: 160px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,111 @@
...
+  margin-top: 19px;
+}
+
+.card__price {
+  font-weight: bold;
+  font-size: 16px;
+  line-height: 19px;
+}
+
+.stars__item:nth-child(-n+4) {
```
vsm2k17, line 87: add classes from stars block task, so if I change the class of your element in dev tools I will see how the count of active stars changing.

src/style.css
```diff
@@ -0,0 +1,111 @@
...
+  text-transform: uppercase;
+  text-decoration: none;
+  background-color: #00acdc;
+  border-radius: 5px;
+}
+
+.card__button:hover {
+  color: #00acdc;
+  background-color: #fff;
+  border: 1px solid #00acdc;
```
vsm2k17, line 110: at this moment when you hovered on your link - border appears and adds 1px, because of box-sizing: border-box; width and height of your button is recalculating, and text could change his position, so its better if you write border property inside the button.

PR #388, opened 2020-07-08
first try

nznmv:
[DEMO LINK](https://nznmv.github.io/layout_product-cards/)
[TEST REPORT LINK](https://nznmv.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,110 @@
+body {
+  margin: 0;
+  font-family: Roboto, sans-serif;
+}
+
+.card {
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+  padding: 30px 14px 14px;
+  margin: 0;
+  max-width: 200px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+
+.card__image {
+  width: 160px;
+  height: 134px;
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,110 @@
...
+  text-decoration: none;
+  text-transform: uppercase;
+  font-weight: bold;
+  font-size: 14px;
+}
+
+.card__button:hover {
+  background-color: white;
+  color: #00acdc;
+  border: 1px solid #00acdc;
```
vsm2k17, line 107: at this moment when you hovered on your link - border appears and adds 1px, because of box-sizing: border-box; width and height of your button is recalculating, and text could change his position, so its better if you write border property inside the button.

src/style.css
```diff
@@ -0,0 +1,110 @@
...
+  box-sizing: border-box;
+  border-radius: 5px;
```
vsm2k17, line 109: no need to repeat this prop, write here only that props which you need to be changed

src/style.css
nznmv: done !

src/style.css
nznmv: ok, done !

PR #461, opened 2020-11-11
solution

xipholena:
- [DEMO LINK](https://xipholena.github.io/layout_product-cards/) 
- [TEST REPORT LINK](https://xipholena.github.io/layout_product-cards/report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,135 @@
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: "Roboto", sans-serif;
+}
+
+body {
+  margin: 0;
+}
+
+.card {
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,135 @@
...
+.card {
+  width: 200px;
+  background: #fff;
+  border-radius: 5px;
+
+  font-weight: normal;
+  font-size: 11px;
+}
+
+.card__inner-width {
```
vsm2k17, line 22: its looks like a modifier of block "card" if that so, it should be called as "card--inner-width"

vsm2k17 (approved):
good job

PR #581, opened 2021-03-30
solution by Dima Chernenko

ched88xtc:
- [DEMO LINK](https://ched88xtc.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://ched88xtc.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,95 @@
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: Roboto, sans-serif;
+  font-weight: 400;
+  font-size: 11px;
+}
+
+body,
+p {
+  margin: 0;
+  padding: 0;
+}
+
+.main__card {
+  width: 200px;
+  padding: 14px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+
+.product-image {
+  height: 134px;
+  width: 160px;
+  margin: 16px 5px 45px;
+  display: flex;
+  justify-content: center;
+}
+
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,95 @@
...
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin-bottom: 20px;
+}
+
+.main__product-stars {
+  display: flex;
+  justify-content: space-between;
+  flex-direction: row;
```
ViPiven, line 53: redundant

ViPiven (approved):

src/style.css
ViPiven, line 53: redundant

vsm2k17: it is redundant because  flex-direction: row is default

src/style.css
```diff
@@ -0,0 +1,95 @@
...
+  text-decoration: none;
+  color: #fff;
+  background-color: #00acdc;
+  border-radius: 5px;
+}
+
+.main__button:hover {
+  color: #00acdc;
+  background-color: #fff;
+  border: 1px solid #00acdc;
```
vsm2k17, line 94: at this moment when you hovered on your link - the border appears and adds 1px, because of box-sizing: border-box; width and height of your button is recalculating, and text could change its position, so it's better if you write border property inside the button.

vsm2k17 (changes requested):

vsm2k17 (approved):
great , fix indentation

PR #606, opened 2021-03-31
Develop

MinchukO:
- [DEMO LINK](https://minchuko.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://minchuko.github.io/layout_product-cards/report/html_report/)

не понятно почему при тестах карточка расширяется
на компьютере всё норм https://prnt.sc/110xkny

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,122 @@
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: "Roboto", sans-serif;
+  font-size: 11px;
+  line-height: 13px;
+  font-weight: 400;
+}
+
+body,
+h2 {
+  margin: 0;
+  --color-button: #00acdc;
+}
+
+/* #product-card */
+
+.page-content__card {
+  max-width: 200px;
+  margin: 0 auto;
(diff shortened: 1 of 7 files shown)


src/style.css
```diff
@@ -0,0 +1,124 @@
...
+/* #button */
+
+.product-card__link {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  background: var(--color-button);
+  border: 1px solid var(--color-button);
+  border-radius: 5px;
+  width: 170px;
```
RuslanaPodshyvalova, line 112: you don't need to set up fixed width: 170px; it is wrong approuch to set up sizes for element here

please set up indents from the left and right sides of your element in 15px; according to the task conditions
  margin-left: 15px;
  margin-right: 15px;

you have alredy set up display: flex; for .product-card__link  so now this block is not inline element but block type element so it spreads its width automaticaly 

you can also define the common css class for all elements on your card in indent 15px and to apply to all of them at once

RuslanaPodshyvalova (changes requested):

PR #750, opened 2021-07-08
Develop

ihor-jpeg:
- [DEMO LINK](https://ihor-jpeg.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://ihor-jpeg.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,100 @@
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap");
+
+html {
+  font-family: Roboto, sans-serif;
+  font-size: 10px;
+  color: #060b35;
+}
+
+body {
+  margin-left: 0;
+}
+
+.card_container {
+  width: 200px;
+  padding: 32px 16px 16px;
+  border: 1px solid #f3f3f3;
+  box-sizing: border-box;
+  border-radius: 5px;
+}
+
+.card_image {
+  display: block;
+  width: 160px;
+  height: 134px;
+  margin: auto;
+}
+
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,100 @@
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap");
+
+html {
+  font-family: Roboto, sans-serif;
+  font-size: 10px;
+  color: #060b35;
```
alex-glotov67, line 6: Maybe use css variable for common colors?

src/style.css
```diff
@@ -0,0 +1,100 @@
...
+  text-decoration: none;
+  text-transform: uppercase;
+  text-align: center;
+  line-height: 40px;
+
+  background-color: #00acdc;
+  border-radius: 5px;
+}
+
+.card_buy-item-button:hover {
```
alex-glotov67, line 94: But in Figma for hover you should add border: 1px solid #0acdc;

alex-glotov67 (changes requested):
All great, but I think you should add a border for button hover

src/style.css
ihor-jpeg: yeah, i don't know how i missed a thing that basic haha. nice catch

src/style.css
ihor-jpeg: yes, great idea. i have fixed that

alex-glotov67 (approved):
Right now it looks perfect

VinnieJ-2k20 (approved):
Good job 👍

PR #760, opened 2021-07-09
BEM try

lera-yaskevych:
- [DEMO LINK](https://lera-yaskevych.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://lera-yaskevych.github.io/layout_product-cards/report/html_report/)

I feel like something is wrong with the placement of margin-top property, but I failed to move it to a proper CSS selector. When I was trying to apply mix (adding another class to the element), bemlint kept saying "error  Element isn't inside the parent block [card__stars] (element-inside-parent-block)". What should be done to this code? Is it OK as it is, or there is a way of making it work better?

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -22,19 +22,17 @@
       >
 
       <div class="card__product-name">
-        <h2 class="card__title">Моноблок APPLE A1419 iMac 27" Retina 5K (MNED2UA/A)</h2>
-        <p class="card__id">Код товара: 195434</p>
+        <h2 class="title card__title">Моноблок APPLE A1419 iMac 27" Retina 5K (MNED2UA/A)</h2>
+        <p class="id card__id">Код товара: 195434</p>
       </div>
 
       <div class="card__reviews">
-        <div class="card__stars">
-          <div class="card__stars card__stars--stars-4">
-            <span class="star"></span>
-            <span class="star"></span>
-            <span class="star"></span>
-            <span class="star"></span>
-            <span class="star"></span>
-          </div>
+        <div class="card__stars stars stars--4">
+          <span class="stars__star"></span>
+          <span class="stars__star"></span>
+          <span class="stars__star"></span>
+          <span class="stars__star"></span>
+          <span class="stars__star"></span>
         </div>
         <p class="card__reviews-count">Отзывов: 5</p>
       </div>
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,56 @@
 <!doctype html>
-<html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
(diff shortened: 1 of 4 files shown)


ihor-jpeg:
After watching BEM videos, I have developed marginophobia for a while. I think margins are okay as long as they are not applied to a block. In your case you don't have any blocks inside the "card" block, so you can only reuse the entire card. I think you can reuse the "buy button", but for that you need to get rid of its margin-top property and create space between the price and the button using padding of your "card__price" element.

lera-yaskevych:
> After watching BEM videos, I have developed marginophobia for a while. I think margins are okay as long as they are not applied to a block. In your case you don't have any blocks inside the "card" block, so you can only reuse the entire card. I think you can reuse the "buy button", but for that you need to get rid of its margin-top property and create space between the price and the button using padding of your "card__price" element.

Good point! Thanks!

src/index.html
```diff
@@ -1,13 +1,48 @@
...
+            user-scalable=no,
+            initial-scale=1.0,
+            maximum-scale=1.0,
+            minimum-scale=1.0
+          ">
+    <title>Product cards</title>
+    <link rel="stylesheet" href="./style.css">
+  </head>
+  <body>
+    <section class="card" data-qa="card">
```
Ihor-Karpyn, line 17:
```suggestion
    <div class="card" data-qa="card">
```

src/index.html
```diff
@@ -1,13 +1,48 @@
...
+      <img
+        src="images/imac.jpeg"
+        alt="photo"
+        class="card__photo"
+      >
+      <h2 class="card__title">Моноблок APPLE A1419 iMac 27" Retina 5K (MNED2UA/A)</h2>
```
Ihor-Karpyn, line 23:
```suggestion
      <h2 class="card__title">
        Моноблок APPLE A1419 iMac 27" Retina 5K (MNED2UA/A)
      </h2>
```

src/style.css
```diff
@@ -0,0 +1,134 @@
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap");
+
```
Ihor-Karpyn, line 2:
```
* {
 box-sizing: border-box;
}
```

src/style.css
```diff
@@ -0,0 +1,134 @@
...
+  --blue-accent: #00acdc;
+}
+
+.card {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  width: 200px;
+  padding: 32px 16px 16px;
+  box-sizing: border-box;
```
Ihor-Karpyn, line 26:
```suggestion
```

Ihor-Karpyn (approved):
Good job🔥

PR #959, opened 2021-10-16
Product cards v1.0

andriibulbuk:
- [DEMO LINK](https://kurtkobein.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://kurtkobein.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/card.css
+++ b/src/card.css
@@ -1,6 +1,4 @@
 .card {
-  font-family: "Roboto", sans-serif;
-  font-style: normal;
   font-size: 12px;
   line-height: 18px;
   color: #060b35;
@@ -14,9 +12,6 @@
 
   display: flex;
   flex-direction: column;
-  
-  --font-weight-l: 700;
-  --font-weight-m: 500;
 }
 
 .card__image {
--- a/src/index.html
+++ b/src/index.html
@@ -26,7 +26,7 @@
       href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet"
     >
   </head>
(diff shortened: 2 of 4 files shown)


src/card.css
```diff
@@ -0,0 +1,90 @@
+.card {
+  font-family: "Roboto", sans-serif;
+  font-style: normal;
```
Val-space, line 3: i think it's better to use in the html file
and font-style: normal;  goes by default, so it's useless

src/card.css
```diff
@@ -0,0 +1,90 @@
...
+  border-radius: 5px;
+
+  width: 200px;
+  padding: 32px 16px 16px;
+
+  display: flex;
+  flex-direction: column;
+  
+  --font-weight-l: 700;
+  --font-weight-m: 500;
```
ihor-jpeg, line 19: use a separate file for variables if you use them outside of this block. (which you do)

src/index.html
```diff
@@ -1,13 +1,69 @@
...
+          <div class="stars__star stars__star--4"></div>
+          <div class="stars__star stars__star--5"></div>
+        </div>
+        <div class="card__reviews">
+          Отзывов: 5
+        </div>
+      </div>
+
+      <div class="card__product-price">
+        <div class="card__price">Цена:</div>
```
ihor-jpeg, line 57:
```suggestion
        <div class="card__price">
          Цена:
        </div>
```

ihor-jpeg (changes requested):
Good work. Just apply those few suggested fixes.

PR #967, opened 2021-10-18
Develop Product Card

darynadudar:
- [DEMO LINK](https://DudarDaryna.github.io/layout_product-cards/)

- [TEST REPORT LINK](https://DudarDaryna.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -12,6 +12,12 @@
     ">
     <title>Product cards</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+    >
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
   </head>
 
   <body>


dmtrhrytsak (approved):
🔥

src/index.html
```diff
@@ -1,13 +1,67 @@
...
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="
+            width=device-width,
+            user-scalable=no,
+            initial-scale=1.0,
+            maximum-scale=1.0,
+            minimum-scale=1.0
+    ">
```
ihor-jpeg, line 12: it is not really a mistake, but i just want to leave this here: the closing symbol should have the same indentation as the opening one. thats the general rule

```suggestion
    <meta name="viewport"
          content="
            width=device-width,
            user-scalable=no,
            initial-scale=1.0,
            maximum-scale=1.0,
            minimum-scale=1.0,
        "
    >
```

ihor-jpeg: but again, here you only have two attributes. you should start putting attributes each on its own line after their number is >= 3

src/index.html
```diff
@@ -1,13 +1,67 @@
...
+    <title>Product cards</title>
+    <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+    >
```
ihor-jpeg, line 19: either do it everywhere, or nowhere. if you only have 2 attributes, then i suggest nowhere :)

src/index.html
```diff
@@ -1,13 +1,67 @@
...
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+  </head>
+
+  <body>
+    <div class="product-card" data-qa="card">
+      <img
+        src="images/imac.jpeg"
+        alt=""
+        class="product-card__product-img"
+      >
```
ihor-jpeg, line 29: this is correct 👍

src/index.html
```diff
@@ -1,13 +1,67 @@
...
+        </div>
+
+        <span class="product-card__reviews-count">
+          Отзывов: 5
+        </span>
+      </div>
+
+      <div class="product-card__price price">
+        <span class="price__title">Цена:</span>
+        <span class="price__value">69 999 грн</span>
```
ihor-jpeg, line 56: have a look. in the code above you formatted it correctly. do it with the rest.
```suggestion
        <span class="price__title">
            Цена:
        </span>
        <span class="price__value">
            69 999 грн
        </span>
```

src/index.html
```diff
@@ -1,13 +1,67 @@
...
+      </div>
+
+      <a href="#"
+        class="product-card__buy-button"
+        data-qa="hover"
+      >
```
ihor-jpeg, line 62:
```suggestion
      <a
        href="#"
        class="product-card__buy-button"
        data-qa="hover"
      >
```

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+}
+
+.product-card__product-name {
+  font-size: 12px;
+  line-height: var(--line-height-big);
+  font-weight: 500;
+  margin-bottom: 4px;
+}
+
+.product-card__product-number {
```
ihor-jpeg, line 44: it helps a lot to have variables and classes with descriptive names. there can be many numbers on a card and each with a different purpose. would make more sense to name it `product-id` or something

src/style.css
```diff
@@ -0,0 +1,106 @@
...
+  line-height: 16px;
+  background: var(--blue-accent);
+  border: 1px solid transparent;
+  border-radius: 5px;
+}
+
+.product-card__buy-button:hover {
+  color: var(--blue-accent);
+  background: #fff;
+  border: 1px solid var(--blue-accent);
```
ihor-jpeg, line 105: you button already has a border. it is just not visible. don't create a new border here. would make more sense to only change `border-color`.

src/index.html
DudarDaryna: https://prnt.sc/1wpif5x

So, what I shoul do? Leave this line or formating? I don't really understand

PR #970, opened 2021-10-18
Cards

Veronika-Korona:
- [DEMO LINK](https://veronika-korona.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://veronika-korona.github.io/layout_product-cards/report/html_report/)

Возможно стоило подключить стили css несколькими отдельными файлами?

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,132 @@
+/* #region Normalize */
+html {
+  font-family: Roboto, Helvetica, sans-serif;
+  font-size: 10px;
+  font-weight: 400;
+  color: #060b35;
+}
+
+body {
+  margin: 0;
+  box-sizing: border-box;
+  width: 200px;
+}
+/* #endregion */
+
+/* #region Product card */
+.product-card {
+  padding: 32px 16px 16px;
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  background: #fff;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+/* #endregion */
+
+.product-card__image {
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,132 @@
+/* #region Normalize */
+html {
+  font-family: Roboto, Helvetica, sans-serif;
+  font-size: 10px;
+  font-weight: 400;
+  color: #060b35;
+}
```
alinaShchavinska, line 7: it may be better to add this part in the new reset.css file?

alinaShchavinska (approved):
Good work

src/style.css
```diff
@@ -0,0 +1,132 @@
...
+.product-card__image {
+  width: 160px;
+  height: 134px;
+  margin: 0 auto 40px;
+}
+
+/* #region Product info */
+.product-card__data {
+  margin-top: 0;
+  display: inline-block;
```
ihor-jpeg, line 37: `display: inline-block;` <--- but why?

src/style.css
```diff
@@ -0,0 +1,132 @@
...
+  font-weight: 700;
+  font-size: 16px;
+  text-align: right;
+}
+/* #endregion */
+
+/* #region Link */
+.product-card__button {
+  box-sizing: border-box;
+  width: 100%;
```
ihor-jpeg, line 114: no need for this. your button is inside a flex container. its display property is set to block. so it is 100% by default

ihor-jpeg (approved):
Good work! you can create a new file per block. you can prob do it in your next assignment

PR #976, opened 2021-10-19
Product card was added

Vitalii-Lytvyn01:
[DEMO LINK](https://Vitalii-Lytvyn01.github.io/layout_product-cards/)
[TEST REPORT LINK](https://Vitalii-Lytvyn01.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,121 @@
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap");
+
+* {
+  box-sizing: border-box;
+}
+
+html {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,121 @@
...
+}
+
+.card__tag {
+  color: #616070;
+  line-height: 18px;
+}
+
+.card__buy-button {
+  height: 40px;
+  width: 100%;
```
ihor-jpeg, line 77: width: 100% is not needed here, because your card is a flex container and the button is its direct child. It will behave like a block element

PR #983, opened 2021-10-19
Develop

Nadiyahr:
- [DEMO LINK](https://Nadiyahr.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://Nadiyahr.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/styles/card.css
+++ b/src/styles/card.css
@@ -59,7 +59,6 @@
 
 .card__button {
   background-color: var(--blue-color);
-  /* width: 100%; */
   text-align: center;
   line-height: 14px;
   padding: 12px 0;


src/styles/reset.css
```diff
@@ -0,0 +1,16 @@
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+}
```
msaikun, line 8: Here you don`t reset the styles. You can move this code to src/style.css.

src/styles/variables.css
```diff
@@ -0,0 +1,6 @@
+:root {
+  --button-color: #00acdc;
+  --border-color: #f3f3f3;
+  --dark-font: #060b35;
+  --light-font: #616070;
```
msaikun, line 5: You use bad names for your variables. Come up with something else.

src/index.html
```diff
@@ -1,13 +1,71 @@
...
+          initial-scale=1.0,
+          maximum-scale=1.0,
+          minimum-scale=1.0
+        ">
+    <title>Product cards</title>
+    <link rel="stylesheet" href="./style.css">
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
+      rel="stylesheet"
+    >
```
ihor-jpeg, line 18: only two attributes. it is unnecessary

src/styles/card.css
```diff
@@ -0,0 +1,82 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  width: 200px;
+  padding: 32px 16px 16px;
+  justify-content: space-evenly;
```
ihor-jpeg, line 6: this doesn't really do anything

src/styles/card.css
```diff
@@ -0,0 +1,82 @@
...
+}
+
+.card__img {
+  width: 160px;
+  height: 134px;
+  margin: 0 auto 40px;
+}
+
+.card__name {
+  display: inline-block;
```
ihor-jpeg, line 19: why would you need to make it an inline-block? it does not seem that you set the width or height
write less code

src/styles/card.css
```diff
@@ -0,0 +1,82 @@
...
+
+.card__price {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin-bottom: 12px;
+}
+
+.card__price-title {
+  display: inline-block;
```
ihor-jpeg, line 51: but why?

src/styles/card.css
```diff
@@ -0,0 +1,82 @@
...
+  color: var(--light-color);
+  line-height: 18px;
+}
+
+.card__price-sum {
+  display: inline-block;
```
ihor-jpeg, line 57: but-why-wtf

src/styles/card.css
```diff
@@ -0,0 +1,82 @@
...
+  font-size: 16px;
+  font-weight: 700;
+  line-height: 18px;
+  padding-bottom: 4px;
+}
+
+.card__button {
+  background-color: var(--blue-color);
+  width: 100%;
```
ihor-jpeg, line 66: write less code. it is already 100% because its display prop. is set to block, because its parent is a flex conntainer

ihor-jpeg (changes requested):
Good work, but needs a tiny bit more work 👍

src/styles/card.css
```diff
@@ -0,0 +1,78 @@
...
+.card__price-sum {
+  font-size: 16px;
+  font-weight: 700;
+  line-height: 18px;
+  padding-bottom: 4px;
+}
+
+.card__button {
+  background-color: var(--blue-color);
+  /* width: 100%; */
```
ihor-jpeg, line 62: do not leave commented code

PR #1314, opened 2022-04-05
product card

valerii-derkach:
- [DEMO LINK](https://valerii-derkach.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://valerii-derkach.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,44 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <title>Product cards</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Product cards</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <title>Product cards</title>
+    <link rel="stylesheet" href="./styles/index.css">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+  </head>
+
+  <body class="page">
+    <div class="card page__card" data-qa="card">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -1,13 +1,44 @@
...
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <title>Product cards</title>
+    <link rel="stylesheet" href="./styles/index.css">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+  </head>
+
+  <body class="page">
+    <div class="card page__card" data-qa="card">
+      <img src="./images/imac.jpeg" alt="imac image" class="card__image">
```
iwanttosmile4u, line 14: Please, write attributes from a new line. It'll be better to read them.

src/index.html
```diff
@@ -1,13 +1,44 @@
...
+        </div>
+        <span class="card__reviews">Отзывов: 5</span>
+
+      </div>
+      <span class="card__price-container">
+        <div class="card__price">Цена:</div>
+        <span class="card__amount">69 999 грн</span>
+
+      </span>
+      <a class="card__buy" href="#buy" data-qa="hover">
```
iwanttosmile4u, line 38: In this case also will be great if you write each attribute from a new line.

src/styles/reset.css
```diff
@@ -0,0 +1,15 @@
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
```
iwanttosmile4u, line 7: I think this rule is redundant, because you have different sizes for font. You duplicate this rule in different classe - card__description, for example. But it's my suggestion only.

src/index.html
valerii-derkach: I agree, changed

src/index.html
valerii-derkach: I think it could be left here, but according to the case above, I corrected it to the same view, as you remarked

src/styles/reset.css
valerii-derkach: yes, a good remark, i have removed font-size from html, every block has it own font-size

Anna-Tkachenko (approved):
Well Done! ✔️

PR #1349, opened 2022-05-05
all test passed

Galyopa:
- [DEMO LINK](https://galyopa.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://galyopa.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Product cards
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_product-cards/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_product-cards/report/html_report/)
+- [DEMO LINK](https://galyopa.github.io/layout_product-cards/)
+- [TEST REPORT LINK](https://galyopa.github.io/layout_product-cards/report/html_report/)
 
 > Follow [this instructions](https://mate-academy.github.io/layout_task-guideline)
 ___
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,60 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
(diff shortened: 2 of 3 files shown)


src/styles/reset.css
```diff
@@ -0,0 +1,14 @@
+body,
+h2 {
+  margin: 0;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+h2 {
```
mgrinko, line 10: I think it is better to have all the styles for h2 together
Because is it not only reset any more you could name the file `normalize.css`

src/styles/reset.css
```diff
@@ -0,0 +1,14 @@
+body,
+h2 {
```
mgrinko, line 2: In Colored blocks solution I explained why it is not a good idea to spread styles across multiple rule sets

src/styles/stars.css
```diff
@@ -0,0 +1,21 @@
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url("../images/star.svg");
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
```
mgrinko, line 9: http://codeguide.academy/html-css.html#css-order

src/styles/style.css
```diff
@@ -0,0 +1,87 @@
...
+
+html {
+  font-family: "Roboto", sans-serif;
+  font-weight: 400;
+  font-size: 12px;
+  line-height: 18px;
+  color: var(--black-text);
+}
+
+.product-card {
```
mgrinko, line 13: Block styles should be in a separate file

src/styles/style.css
```diff
@@ -0,0 +1,87 @@
...
+.product-card__rate {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin-bottom: 24px;
+}
+
+.product-card__rewies {
+  font-size: 10px;
+  line-height: 14px;
```
mgrinko, line 52: It is the second time you used this size and lh
Maybe it is the default one?

mgrinko (approved):

src/styles/style.css
Galyopa: I have to move html tag into normalize.css?

PR #1381, opened 2022-05-10
commit message

Bohdan-Pyschchyk:
- [DEMO LINK](https://Bohdan-Pyschchyk.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://Bohdan-Pyschchyk.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,152 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.card {
+  max-width: 200px;
+  padding: 15px 16px;
+
+  font-family: "Roboto", sans-serif;
+
+  border-radius: 5px;
+  border: 1px solid #f3f3f3;
+  background: #fff;
+}
+
+.card__image-link {
+  display: block;
+  margin: 17px 3px 0;
+}
+
+.card__image {
+  display: block;
+  width: 100%;
+  height: 134px;
+  text-align: center;
+}
+
+.card__info-name {
+  margin: 40px 0 0 ;
+
+  text-align: left;
+  font-weight: 500;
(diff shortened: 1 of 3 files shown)


yanaost (approved):
you could use variables for 2 main colors (you have used 060b35 6 times) and normalize/reset.css

src/Styles/index.css
```diff
@@ -0,0 +1,92 @@
+@import "./star.css";
+@import "./button.css";
+
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
```
jstmpelowycz, line 10: That's better to spread these styles over several files according to their dedication

src/Styles/index.css
```diff
@@ -0,0 +1,92 @@
...
+
+.card {
```
jstmpelowycz, line 12: And styles with the prefix `card__` should be kept in a file named `card.css` (as of BEM)

src/Styles/index.css
```diff
@@ -0,0 +1,92 @@
...
+  margin: 0;
+
+  font-weight: 400;
+  font-size: 12px;
+  line-height: 18px;
+  color: #616070;
+}
+
+.card__price-tag {
+  margin: 0;
```
jstmpelowycz, line 86: That's better to reset margins (or anything else) using resetting styles, not here

src/style.css
```diff
@@ -0,0 +1,152 @@
+* {
```
jstmpelowycz, line 1: You don't need this file anymore

jstmpelowycz (approved):
Good job, but make sure you check out the comments left down below

PR #1624, opened 2022-07-16
Add product card

VolodymyrChuyko:
- [DEMO LINK](https://VolodymyrChuyko.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://VolodymyrChuyko.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,62 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <title>Product cards</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Product cards</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1.0"
+    >
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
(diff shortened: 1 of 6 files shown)


OlesiaNikishaeva (approved):

src/index.html
```diff
@@ -1,13 +1,62 @@
...
+  </head>
+  <body>
+    <div class="card" data-qa="card">
+      <img
+        class="card__image"
+        src="images/imac.jpeg"
+        alt="product image"
+      >
+
+      <h2 class="card__header">
```
dellachaise, line 23: `card__title` - better name for this class, because you use it on `h2` tag, not `header`.

src/styles/card.css
```diff
@@ -0,0 +1,78 @@
...
+  padding: 11px 0;
+  font-weight: 700;
+  font-size: 14px;
+  line-height: 16px;
+  color: #fff;
+  text-align: center;
+  text-transform: uppercase;
+  text-decoration: none;
+  background-color: #00acdc;
+  border: 1px solid #00acdc;
```
dellachaise, line 70: Better use `transparent` border color this way.

dellachaise (changes requested):
Good job 👍 

Please, add  2 little fixes according to comments.

src/index.html
```diff
@@ -13,14 +13,14 @@
     <title>Product cards</title>
   </head>
   <body>
-    <div class="card" data-qa="card">
+    <article class="card" data-qa="card">
```
dellachaise, line 16: I don't think that `article` the best tag for this block.

It's more about text articles, more you could read [there](https://developer.mozilla.org/ru/docs/Web/HTML/Element/article)

VolodymyrChuyko: I changed it to `article` because I suppose it would fit better for the layout_catalog. There I have a `section`  as a block of cards. Cards are `articles` that contain `p` tags

PR #1697, opened 2022-07-29
Product-card done

Oleksii-Blyzniuk:
(https://Oleksii-Blyzniuk.github.io/layout_product-cards/)
(https://Oleksii-Blyzniuk.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,123 @@
+body {
+  margin: 0;
+  box-sizing: border-box;
+  font-family: "Roboto", Arial, sans-serif;
+  color: #050b35;
+}
+
+.container {
+  width: 200px;
+}
+
+.card {
+  background: #fff;
+  border-radius: 5px;
+  display: flex;
+  align-items: center;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,123 @@
...
+  margin: 0;
+  box-sizing: border-box;
+  font-family: "Roboto", Arial, sans-serif;
+  color: #050b35;
+}
+
+.container {
+  width: 200px;
+}
+
```
valerii-derkach, line 11: do not set sizes to blocks, use mix where necessary, and move all external styles under element selector.

valerii-derkach (changes requested):
You should create separate file for each BEM block

ArturErmolenko (changes requested):
Fix button size, your button is bigger that in mockup,need to fix it

alexandra-protyanova (approved):
Well done!

valerii-derkach (approved):

PR #1712, opened 2022-08-05
finished task

Nataliia95254:


src/style.css
```diff
@@ -0,0 +1,121 @@
...
+  padding: 0;
+}
+
+body {
+  font-family: "Roboto", sans-serif;
+}
+
+a {
+  text-decoration: none;
+}
```
Bohdan-Maksymiuk, line 14: use class selectors

src/style.css
```diff
@@ -0,0 +1,121 @@
...
+
+.card__amount {
+  font-size: 16px;
+  font-weight: 700;
+  line-height: 18px;
+  color: #060b35;
+}
+
+.card__button-buy {
+  width: 166px;
```
Bohdan-Maksymiuk, line 105: it should be 100%

src/style.css
```diff
@@ -0,0 +1,121 @@
...
+  text-decoration: none;
+}
+
+.card {
+  box-sizing: border-box;
+  padding: 32px 16px 16px;
+  display: flex;
+  flex-direction: column;
+  width: 200px;
+  height: 400px;
```
Bohdan-Maksymiuk, line 22: do not hardcode height - it should be stretched by content

Bohdan-Maksymiuk (changes requested):

ArturErmolenko (approved):

PR #1719, opened 2022-08-06
Product card was created

mykyta-krainik:
https://mykyta-krainik.github.io/layout_product-cards/
https://mykyta-krainik.github.io/layout_product-cards/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,65 @@
 <!doctype html>
-<html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <title>Product cards</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Product cards</h1>
-</body>
+<html lang="en" class="page">
+  <head>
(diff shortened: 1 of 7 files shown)


src/index.html
```diff
@@ -1,13 +1,65 @@
...
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </span>
+
+        <span class="card__product-code card__product-code--wrapper">
+          Product code: 195434
+        </span>
+
+        <div class="card__rating card__rating--wrapper">
+          <div class="stars stars--4">
+            <div class="stars__star stars__star--wrapper"></div>
```
alexandra-protyanova, line 39: why do u need second class & it's definately not wrapper

alexandra-protyanova (changes requested):

src/index.html
mykyta-krainik: Second class is needed to make gaps between stars, because I want to follow the BEM methodology.
Hmm, I think so. Perhaps the name like "stars__star--right-gap_4px" is more suitable in this case

alexandra-protyanova: stars__star isn't block, it is element so u can set gaps there

alexandra-protyanova (changes requested):
Remove this class)

alexandra-protyanova (approved):
well done 👍

PR #1751, opened 2022-08-16
add card

AndriyZhabynets:


src/card.css
```diff
@@ -0,0 +1,88 @@
...
+  color: #616070;
+}
+
+.card__price-total {
+  font-weight: 700;
+  font-size: 16px;
+  line-height: 18px;
+}
+
+.card__buy-bottom {
```
imaksweb, line 66: According to the checklist you need to add a transition for your buy button hover effect

src/index.html
```diff
@@ -1,13 +1,61 @@
...
+  </head>
+  <body>
+    <div class="card" data-qa="card">
+      <img
+        src="./images/imac.jpeg"
+        class="card__image"
+        alt="image"
+      >
+
+      <h1 class="card__title">
```
imaksweb, line 21: I'm not sure that it's a good idea to use h1 here, h1 should be one in a page. But I think it's not critical in this task

src/index.html
```diff
@@ -1,13 +1,61 @@
...
+        </div>
+        <p>
+          Reviews: 5
+        </p>
+      </div>
+
+      <div class="card__price">
+        <h3 class="card__price-text">
+          Price:
+        </h3>
```
imaksweb, line 45: I suppose you can use p tag instead of h3

src/card.css
```diff
@@ -0,0 +1,88 @@
...
+.card__title {
+  margin: 0 16px 4px;
+  line-height: 18px;
+  font-size: 12px;
+  font-weight: 500;
+  color: #060b35;
+}
+
+.card__product-code {
+  margin-left: 16px;
```
Anna-Tkachenko, line 26: Let's specify correct general padding to the whole card instead of adding margin-left / right for all elements

Anna-Tkachenko (changes requested):
Good job!
Just need one small fix with margins-paggings)

src/card.css
AndriyZhabynets: Ok

Anna-Tkachenko (approved):
Good job! 👍

PR #1752, opened 2022-08-16
solution

vmamchur:
[DEMO LINK](https://<your_account>.github.io/layout_product-cards/)
[TEST REPORT LINK](https://<your_account>.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/styles/blocks/card.css
+++ b/src/styles/blocks/card.css
@@ -1,0 +1,112 @@
+.card {
+  font-family: Roboto, sans-serif;
+
+  display: flex;
+  flex-direction: column;
+
+  width: 200px;
+
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+
+  padding: 32px 16px 16px;
+}
+
+.card__image {
+  width: 160px;
+  height: 134px;
+
+  align-self: center;
+
+  margin: 0 0 40px;
+}
+
+.card__title {
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 18px;
+
+  color: #060b35;
+
+  margin: 0 0 4px;
+}
+
+.card__code {
+  font-size: 10px;
+  line-height: 14px;
+
+  color: #616070;
+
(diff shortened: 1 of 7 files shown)


src/styles/blocks/card.css
```diff
@@ -0,0 +1,113 @@
...
+    color 0.3s;
+}
+
+.card__link:hover {
+  background-color: #fff;
+
+  color: var(--color-button);
+
+  border-color: var(--color-button);
+  border-radius: var(--border-radius-button);
```
artem-musii, line 112: You have border-radius in card__link, so maybe you don't need to add it here and you also could set border-color in card__link, so you wouldn't have to change it here

artem-musii (commented):
Everything looks fine for me, but maybe you didn't need to change border styles in card__link:hover

src/styles/blocks/card.css
```diff
@@ -0,0 +1,110 @@
+* {
+  box-sizing: border-box;
+}
```
Denchepornyuk, line 3: add this selector in index.css file

src/styles/blocks/card.css
```diff
@@ -0,0 +1,110 @@
...
+
+.card {
+  font-family: Roboto, sans-serif;
```
Denchepornyuk, line 6: add font-family to html tag

Denchepornyuk (changes requested):

src/styles/blocks/card.css
```diff
@@ -0,0 +1,104 @@
...
+
+.card__price-value {
+  font-weight: 700;
+  line-height: var(--line-height-big);
+
+  color: var(--color-dark);
+}
+
+.card__link {
+  height: 40px;
```
OleksiiBlezniuk, line 78: As for me, it's better to set paddings instead of height hardcoding

OleksiiBlezniuk (approved):
GJ, left some suggestions about code readability

PR #1769, opened 2022-08-18
Develop

grafit-off:
- [DEMO LINK](https://grafit-off.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://grafit-off.github.io/layout_product-cards/report/html_report/)


src/styles/settings.css
```diff
@@ -0,0 +1,13 @@
+.html {
+  box-sizing: border-box;
+
+  font-family: "Roboto", sans-serif;
+}
+
+* {
+  box-sizing: border-box;
+}
```
Rina07, line 9: Great job! Some small improvements on my opinion would be: 
delete box-sizing in html as it's applied to all elements anyway, and set standard font-size and font-color in html for minimizing code in classes)

Rina07 (approved):

src/styles/settings.css
grafit-off: Why do I use box sizing this way?  Instead of the usual * {...}, because some libraries, according to the Swiper type, use box-sizing: content-box (e.g. the swiper-wrapper class) and for the library to display as as the developer intended, she needs to leave box-sizing, specified by the developer. So I use imitation. 
Set standard font-size and font-color in html for minimizing code in classes ,it's true, will change this now

src/styles/fonts.css
```diff
@@ -0,0 +1,17 @@
...
+  font-family: Roboto;
+  src: url(../fonts/Roboto-Medium.woff);
+  font-weight: 500;
+}
+
+@font-face {
+  font-family: Roboto;
+  src: url(../fonts/Roboto-Bold.woff);
+  font-weight: 700;
+}
```
OleksiiBlezniuk, line 17: Why not google fonts? It will be much easier, and also there will be no need to upload font files

src/styles/settings.css
```diff
@@ -0,0 +1,15 @@
+.html {
+  box-sizing: border-box;
```
OleksiiBlezniuk, line 2: Do `html` have any borders?

OleksiiBlezniuk: I don't think that html should have it

src/index.html
```diff
@@ -1,13 +1,92 @@
...
+
+          <span class="stars__star"></span>
+
+          <span class="stars__star"></span>
+
+          <span class="stars__star"> </span>
+        </div>
+
+        <p class="card__reviews">Reviews: 5</p>
+      </div>
```
OleksiiBlezniuk, line 75:
```suggestion
        <div
          class="card__stars stars stars--4"
          aria-label="Rating of this product is 4 out of 5."
        >
          <span class="stars__star"></span>
          <span class="stars__star"></span>
          <span class="stars__star"></span>
          <span class="stars__star"></span>
          <span class="stars__star"> </span>
        </div>
        <p class="card__reviews">Reviews: 5</p>
      </div>
```

OleksiiBlezniuk: Use blank lines to separate big blocks of code, not for every element

src/styles/components/card.css
```diff
@@ -0,0 +1,95 @@
...
+  width: 160px;
+  height: 134px;
+
+  margin: 0 auto 40px;
+
+  object-fit: contain;
+}
+
+.card__heading {
+  margin: 0 0 4px;
```
OleksiiBlezniuk, line 23: Ok, but I think it would be better for readability to reset default margins somewhere on the top of file (or even in separate file) and set only bottom margin here

OleksiiBlezniuk (approved):

src/index.html
grafit-off: Ok, will change this soon

src/styles/components/card.css
grafit-off: To be honest also think about it, but i dont really like to use tags. But if i create class for reset, he is gonna be block (BEM), but block can't have margins, etc.. So i'm just can't decide which is more correct. Next time will use tag for reset like other.

PR #1781, opened 2022-08-18
product_card added

Yeliseiev:
- [DEMO LINK](https://Yeliseiev.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://Yeliseiev.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,56 @@
 <!doctype html>
-<html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <title>Product cards</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Product cards</h1>
-</body>
+<html lang="en" class="home-page">
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <title>Product cards</title>
+    <link rel="stylesheet" href="./style.css">
+  </head>
+  <body class="home-page__content">
+
+    <div class="card" data-qa="card">
+
+      <img
+          src="./images/imac.jpeg"
+          alt="card__image"
+          class="card__image">
+
+      <div class="card__info">
+        <h2 class="card__info-name">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </h2>
+        <p class="card__info-code">
+          Product code: 195434
+        </p>
+
+        <div class="card__info-reviews">
+          <div class="stars stars--4">
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+          </div>
+            Reviews: 5
+        </div>
+
+        <div class="card__info-price">
+          <div class="card__info-price-title">
+            Price:
+          </div>
+          <div class="card__info-price-value">
+            $2,199
+          </div>
+        </div>
+
+      </div>
+
+      <button class="card__button" data-qa="hover">
+        buy
+      </button>
+    </div>
+
+
+  </body>
 </html>
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Product cards
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_product-cards/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_product-cards/report/html_report/)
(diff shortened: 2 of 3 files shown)


src/reset.css
```diff
@@ -0,0 +1,14 @@
+* {
+  box-sizing: border-box;
+  --primary-text-color: #060b35;
+  --secondary-text-color: #616070;
+  --bg-color: #fff;
+  --button-color: #00acdc;
```
imaksweb, line 6: I'm not sure that variables should be in the reset file

src/reset.css
```diff
@@ -0,0 +1,14 @@
...
+}
+
+.home-page__content {
+  margin: 0;
+  font-family: "Roboto", sans-serif;
+  font-weight: 400;
+  color: var(--primary-text-color);
```
imaksweb, line 13: I think it's also shouldn't be there

src/card.css
```diff
@@ -0,0 +1,93 @@
...
+  width: 200px;
+
+  background: var(--bg-color);
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+
+.card__image {
+  width: 160px;
+  height: 134px;
```
imaksweb, line 15: Probably it's not a good idea to hardcode height, I think just use width would be great

src/card.css
```diff
@@ -0,0 +1,93 @@
...
+  margin: 0 auto;
+}
+
+.card__info {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  padding-top: 40px;
+  padding-bottom: 16px;
```
imaksweb, line 24: If you have paddings or margins for more than 1 side better to use this note padding: 40px 0 16px

src/card.css
```diff
@@ -0,0 +1,93 @@
...
+  padding-top: 4px;
+  padding-bottom: 16px;
+}
+
+.card__info-reviews {
+  display: flex;
+  justify-content: space-between;
+
+  font-size: 10px;
+  line-height: 14px;
```
imaksweb, line 50: I think you can remove this line, it's completely no sense. I guess you just copied it from figma)

src/index.html
```diff
@@ -1,13 +1,60 @@
...
+
+        <div class="card__info-reviews">
+          <div class="stars stars--4">
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+          </div>
+            Reviews: 5
```
imaksweb, line 39: From my perspective div is something about layout so I'd rather wrap this text in a span or paragraph.

src/index.html
```diff
@@ -1,13 +1,60 @@
...
+        </div>
+
+        <div class="card__price">
+          <div class="card__price-title">
+            Price:
+          </div>
+          <div class="card__price-value">
+            $2,199
+          </div>
```
imaksweb, line 48: The same thing I think it's better to use paragraph instead of div

src/index.html
```diff
@@ -1,13 +1,60 @@
...
+        </div>
+
+      </div>
+
+      <button class="card__button" data-qa="hover">
+        buy
+      </button>
```
imaksweb, line 55: According to checklist you should use a tag, not button

src/card.css
```diff
@@ -0,0 +1,93 @@
...
+
+  background: var(--button-color);
+  color: var(--bg-color);
+  text-transform: uppercase;
+
+  padding: 12px 0;
+
+  border: none;
+  border-radius: 5px;
+  transition: color 0.3s;
```
imaksweb, line 86: You need to add transition for background-color too

src/card.css
Yeliseiev: it's didn't pass tests without height

src/index.html
Yeliseiev: Not sure, because i think <p> has means that it's paragraph, and must starts from new line

src/card.css
imaksweb: Sorry, you're right. I had the same problem)

src/index.html
imaksweb: > Not sure, because i think p has means that it's paragraph, and must starts from new line

Yeah, but you can use flex to place them on a same line

src/card.css
Yeliseiev: Yes I copied it from figma, and it's not working without this line

src/index.html
```diff
@@ -1,13 +1,63 @@
...
+    <link rel="stylesheet" href="./style.css">
+  </head>
+  <body class="home-page__content">
+
+    <div class="card" data-qa="card">
+
+      <img
+          src="./images/imac.jpeg"
+          alt="card__image"
+          class="card__image">
```
Anna-Tkachenko, line 21: Check indents
```suggestion
      <img
        src="./images/imac.jpeg"
        alt="card__image"
        class="card__image"
      >
```

Anna-Tkachenko: And will be better to specify more understandable alt text, imac, for example

Anna-Tkachenko (approved):
Good job! 🔥

src/index.html
Yeliseiev: Yes, sorry I'm always overlook this part.

PR #1792, opened 2022-08-18
Product card was created

Yevheniya-Deryka:
- [DEMO LINK](https://Yevheniya-Deryka.github.io/layout_cards/)
- [TEST REPORT LINK](https://Yevheniya-Deryka.github.io/layout_cards/report/html_report/)

(Тести та деплой не працюють)

### DIFF
--- a/src/card.css
+++ b/src/card.css
@@ -1,0 +1,88 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  max-width: min-content;
+
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+
+  padding: 16px;
+}
+
+.card__content {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
+
+.card__image {
+  width: 160px;
+  height: 134px;
+
+  margin: 16px 3px 0;
+}
+
+.card__title {
+  font-weight: 500;
+  line-height: 18px;
+
+  width: 166px;
+
+  margin-top: 40px;
+}
+
+.card__code {
+  color: #616070;
+
+  margin-top: 4px;
+}
+
+.card__reviews {
+  padding-top: 16px;
+  font-size: 10px;
+}
+
+.card__price {
+  display: inline-block;
+
+  color: #616070;
+
+  padding-top: 24px;
+  margin-left: 0;
+}
+
+.card__price-amount {
+  font-weight: 700;
+  font-size: 16px;
+  line-height: 18px;
+
+  padding-top: 24px;
+}
+
+.card__buy-button {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  width: 166px;
+  height: 40px;
+  margin-top: 16px;
+
+  border: 1px solid transparent;
+  border-radius: 5px;
+  background-color: #00acdc;
+
+  text-transform: uppercase;
+  color: #fff;
+  font-size: 14px;
+  font-weight: 700;
+
+  cursor: pointer;
+  transition: color 0.2s;
+}
+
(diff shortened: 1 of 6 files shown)


Apollonia5 (changes requested):
невеликі зауваження

src/card.css
```diff
@@ -0,0 +1,88 @@
...
+}
+
+.card__price {
+  display: inline-block;
+
+  color: #616070;
+
+  padding-top: 24px;
+  margin-left: 0;
+}
```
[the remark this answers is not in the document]
Yevheniya-Deryka: є) посередині

src/index.html
```diff
@@ -1,13 +1,57 @@
...
+      <a
+        href="/"
+        data-qa="hover"
+        class="card__buy-button"
+      >
+        Buy
+      </a>
+    </div>
+  </body>
 </html>
```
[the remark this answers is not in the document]
Apollonia5: це перша моя перевірка, пробачайте

src/card.css
```diff
@@ -0,0 +1,89 @@
...
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
+
+.card__image {
+  width: 160px;
+  height: 134px;
+
+  margin: 16px 3px 0;
```
OleksiiBlezniuk, line 22: that 3px margins looks strange, you used it for aligning? in that case better set auto or use flex

src/card.css
```diff
@@ -0,0 +1,89 @@
...
+}
+
+.card__title {
+  font-weight: 500;
+  line-height: 18px;
+
+  width: 166px;
```
OleksiiBlezniuk, line 29: Why have you hardcoded title width? It's probably should take all free space

src/card.css
```diff
@@ -0,0 +1,89 @@
...
+  font-size: 16px;
+  line-height: 18px;
+
+  padding-top: 24px;
+}
+
+.card__buy-button {
+  display: block;
+
+  width: 166px;
```
OleksiiBlezniuk, line 66: same here, don't hardcode blocks width, if there is no need

src/card.css
OleksiiBlezniuk: Okay, I got it, you haven't set width to `card`
Instead of setting width for every element, better to set width to parent, and then if there will be need to make block wider - set width: 100% (because due to design there is no options where card can be wider)

Because it's looks strange now, we know that card should have 200px width, but instead of setting it we've calculated how many width should content have and set it to elements

It's great to let content dictate container size instead of hardcoding it, but in this case we have hardcoded it twice 🙂  So let's make it a bit simpler

OleksiiBlezniuk (changes requested):
Good job! There are few small coments
Also would be nice to organize your files, move css files to styles folder, since there are more than one file

src/index.html
```diff
@@ -1,13 +1,57 @@
...
+  </head>
+  <body>
+    <div class="card" data-qa="card">
+      <img
+        src="images/imac.jpeg"
+        alt="Imac illustration"
+        class="card__image"
+      >
+
+      <h2 class="card__title">APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</h2>
```
OleksiiBlezniuk, line 22:
```suggestion
      <h2 class="card__title">
        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
      </h2>
```

OleksiiBlezniuk: Too long line

OleksiiBlezniuk (approved):
Good job!

PR #1795, opened 2022-08-18
Develop

jojnajbude:
[DEMO LINK](https://jojnajbude.github.io/layout_product-cards/)
[TEST REPORT LINK](https://jojnajbude.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/blocks/product-card.css
+++ b/src/blocks/product-card.css
@@ -4,7 +4,7 @@
   flex-direction: column;
 
   box-sizing: border-box;
-  padding: 0 16px;
+  padding: 16px;
   width: 200px;
 
   background: #fff;
@@ -16,11 +16,11 @@
   align-self: center;
   height: 134px;
   width: 160px;
-  margin: 32px 0;
+  margin-top: 16px;
 }
 
 .product-card__name {
-  margin: 8px 0;
+  margin-top: 40px;
 }
 
 .product-card__device-model {
@@ -32,7 +32,6 @@
 
 .product-card__code {
   margin-top: 4px;
-  font-weight: 400;
   font-size: 10px;
   line-height: 14px;
   color: #616070;
@@ -41,11 +40,10 @@
 .product-card__rating {
   display: flex;
   justify-content: space-between;
-  margin: 8px 0;
+  margin-top: 16px;
 }
 
 .product-card__reviews {
-  font-weight: 400;
   font-size: 10px;
   line-height: 14px;
 
@@ -61,6 +59,7 @@
   flex-direction: row;
   justify-content: space-between;
   align-items: center;
+  margin-top: 24px;
 }
 
 .product-card__price-label {
@@ -71,7 +70,6 @@
 }
 
 .product-card__price {
-  margin: 16px 0;
 
   font-weight: 700;
   font-size: 16px;
@@ -92,13 +90,13 @@
 
   display: inline-block;
   padding: 12px 70px;
-  margin-bottom: 16px;
+  margin-top: 16px;
 
   border: 1px solid var(--blue-accent);
   background: var(--blue-accent);
   border-radius: 5px;
 
-  transition: all 0.3s;
+  transition: color 0.3s, background-color 0.3s;
 }
 
 .product-card__buy-link:hover {


src/index.html
```diff
@@ -1,13 +1,69 @@
...
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </h1>
+
+        <h2 class="product-card__code">
+          Product code: 195434
+        </h2>
+      </div>
+
+      <div class="product-card__rating">
+        <div class="product-card__stars stars stars--4">
```
myroslav-diiak, line 40: You don't need "product-card__stars" class

src/style.css
```diff
@@ -0,0 +1,159 @@
+/* #region Reset */
```
myroslav-diiak, line 1: You should create separate file for each BEM block

myroslav-diiak (changes requested):

myroslav-diiak (approved):

src/index.html
```diff
@@ -1,13 +1,69 @@
...
+      <img
+        class="product-card__image"
+        src="images/imac.jpeg"
+        alt="imac"
+      >
+
+      <div class="product-card__name">
+        <h1 class="product-card__device-model">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </h1>
```
OleksiiBlezniuk, line 32: h1 used only once one the page for the main page title, and card will definitely used more than one time, so it better to use h2 or even h3 here

src/index.html
```diff
@@ -1,13 +1,69 @@
...
+
+        <h2 class="product-card__code">
+          Product code: 195434
+        </h2>
```
OleksiiBlezniuk, line 36: It's just information about product code in store, so it's more just p

src/index.html
```diff
@@ -1,13 +1,69 @@
...
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+        </div>
+
+        <h2 class="product-card__reviews">
+          Reviews: 5
+        </h2>
```
OleksiiBlezniuk, line 50: p

src/index.html
```diff
@@ -1,13 +1,69 @@
...
+      </div>
+
+      <div class="product-card__price-section">
+        <h2 class="product-card__price-label">
+          Price:
+        </h2>
+        <h1 class="product-card__price">
+          $2,199
+        </h1>
```
OleksiiBlezniuk, line 59: h1 can't be used twice or more, it's only for main page title

OleksiiBlezniuk: use p for both of this strings

src/index.html
```diff
@@ -1,13 +1,69 @@
...
+      </div>
+
+      <div class="product-card__buy">
+        <a class="product-card__buy-link" data-qa="hover">
+          BUY
+        </a>
+      </div>
```
OleksiiBlezniuk, line 66: I don't think you really deed this container, just set right styles to the link 🙂

src/blocks/product-card.css
```diff
@@ -0,0 +1,113 @@
+/* #region Product Card */
+.product-card {
+  display: flex;
+  flex-direction: column;
+
+  box-sizing: border-box;
```
OleksiiBlezniuk, line 6: Set it to * selector

src/blocks/product-card.css
```diff
@@ -0,0 +1,113 @@
...
+
+  color: var(--main-accent);
+}
+
+.product-card__price-section {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  gap: 60px;
```
OleksiiBlezniuk, line 64: 1 - this gap don't make any sense since you've used `space-beetween`
2 - better to avoid `gap`, it's not supported by some versions of browser (also browser in tests is not supporting it)

OleksiiBlezniuk (changes requested):

src/blocks/product-card.css
jojnajbude: Don't understand this. What do U mean by this?

src/blocks/product-card.css
```diff
@@ -81,11 +80,6 @@
   color: var(--main-accent);
 }
 
-.product-card__buy {
-
-  margin-bottom: 16px;
-}
```
jojnajbude: And delete useless class for link container

src/blocks/product-card.css
OleksiiBlezniuk: You've used `box-sizing` rule more than one time, so it better to set this property to * selector

src/blocks/product-card.css
```diff
@@ -0,0 +1,109 @@
...
+.product-card__device-model {
+  font-size: 12px;
+  font-weight: 500;
+  line-height: 18px;
+  color: var(--main-accent);
+}
+
+.product-card__code {
+  margin-top: 4px;
+  font-weight: 400;
```
OleksiiBlezniuk, line 35: Default value, it can be removed

src/blocks/product-card.css
```diff
@@ -0,0 +1,109 @@
...
+}
+
+.product-card__rating {
+  display: flex;
+  justify-content: space-between;
+  margin: 8px 0;
+}
+
+.product-card__reviews {
+  font-weight: 400;
```
OleksiiBlezniuk, line 48: Default value, it can be removed

src/blocks/product-card.css
```diff
@@ -0,0 +1,109 @@
...
+.product-card__image {
+  align-self: center;
+  height: 134px;
+  width: 160px;
+  margin: 32px 0;
+}
+
+.product-card__name {
+  margin: 8px 0;
+}
```
OleksiiBlezniuk, line 24: Set margins to one side

OleksiiBlezniuk: Screenshot 2022-08-19 at 15 15 13

src/blocks/product-card.css
```diff
@@ -0,0 +1,109 @@
...
+
+  display: inline-block;
+  padding: 12px 70px;
+  margin-bottom: 16px;
+
+  border: 1px solid var(--blue-accent);
+  background: var(--blue-accent);
+  border-radius: 5px;
+
+  transition: all 0.3s;
```
OleksiiBlezniuk, line 101: Better to define which exact properties will be animated, because if you not define it, it will be applied to every property that can be animated

src/blocks/product-card.css
```diff
@@ -0,0 +1,109 @@
...
+
+.product-card__price-label {
+  font-weight: 400;
+  font-size: 12px;
+  line-height: 18px;
+  color: #616070;
+}
+
+.product-card__price {
+  margin: 16px 0;
```
OleksiiBlezniuk, line 74: Same here with margins

OleksiiBlezniuk (changes requested):

src/blocks/product-card.css
```diff
@@ -0,0 +1,107 @@
+/* #region Product Card */
+.product-card {
+  display: flex;
+  flex-direction: column;
+
+  box-sizing: border-box;
```
MaksimusCRS, line 6:
```suggestion
* {
  box-sizing: border-box;
}

.product-card {
  display: flex;
  flex-direction: column;
```

MaksimusCRS (approved):

PR #1796, opened 2022-08-18
product card was created

Oleksandra-Shevchenko:
https://Oleksandra-Shevchenko.github.io/layout_product-cards/
https://Oleksandra-Shevchenko.github.io/layout_product-cards/report/html_report/


src/styles/star.css
```diff
@@ -0,0 +1,16 @@
...
+.stars_star {
+  background: url(../images/star.svg) no-repeat center;
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+}
+
+.stars2 .stars_star:nth-child(-n+4) {
+  background-image: url(../images/star-active.svg);
+}
```
ArturNagaycev, line 16: i think, you didn`t this point of the task.

Reuse stars block from [Stars task](https://github.com/mate-academy/layout_stars) and keep it operational. **If you change stars--4 modifier to stars--2 or other - number of selected stars should change correctly.**

you didn't add styles for other modifiers

src/styles/card.css
```diff
@@ -0,0 +1,89 @@
...
+}
+
+.card__priceNumber {
+  font-size: 16px;
+  font-weight: 600;
+  color: #060b35;
+}
+
+.card__buttom {
+  box-sizing: border-box;
```
ArturNagaycev, line 69: in previous tasks, Denis wrote me to add box-sizing to * tag ))

src/styles/card.css
```diff
@@ -0,0 +1,89 @@
...
+  width: 166px;
+  border: 1px solid #00acdc;
+  text-transform: uppercase;
+  font-weight: 700;
+  font-size: 14px;
+  line-height: 40px;
+}
+
+.card__buttom:hover {
+  border: 1px solid #00acdc;
```
ArturNagaycev, line 85: you can remove this row in :hover

src/index.html
```diff
@@ -1,13 +1,61 @@
...
+      <h2 class="card__item-name">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </h2>
+
+      <div class="card__product-code">
+        Product code: 195434
+      </div>
+
+      <div class="card__rating">
+        <div class="card__star stars2">
```
ArturNagaycev, line 41: your name of class "card__star" looks like the element "star" of the block "card". but it has to be a block "stars" (name in the task)

src/index.html
```diff
@@ -1,13 +1,61 @@
...
+          <div class="stars_star"></div>
+          <div class="stars_star"></div>
+        </div>
+
+        <div class="card__star-reviews">Reviews: 5</div>
+      </div>
+
+      <div class="card__price">
+        <div class="card__priceName">Price:</div>
+        <div class="card__priceNumber">$2,199</div>
```
ArturNagaycev, line 54: you use different rules for name of elements. use one of them

src/index.html
ArturNagaycev, line 41: in "stars2", "2" is a modifier, we have to separate it from block/element with '--'

ArturNagaycev (changes requested):

src/styles/card.css
Oleksandra-Shevchenko: Thanks, i forgot i wrote this in buttom earlier)

src/index.html
```diff
@@ -1,13 +1,61 @@
...
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </h2>
+
+      <div class="card__product-code">
+        Product code: 195434
+      </div>
+
+      <div class="card__rating">
+        <div class="star stars--4">
+          <div class="stars_star"></div>
```
Anna-Tkachenko, line 42: Check correct syntax for block__element naming
```suggestion
          <div class="stars__star"></div>
```

src/styles/card.css
```diff
@@ -0,0 +1,87 @@
...
+  height: 134px;
+  width: 160px;
+  margin: 32px auto 40px;
+}
+
+.card__item-name {
+  line-height: 18px;
+  font-size: 12px;
+  font-weight: 500;
+  margin: 0 16px 4px;
```
Anna-Tkachenko, line 19: Will be better to specify correct padding for the whole card, instead of left-right margins for all elements, fix it

src/styles/card.css
```diff
@@ -0,0 +1,87 @@
...
+  line-height: 14px;
+  margin-left: 16px;
+}
+
+.card__price {
+  display: flex;
+  justify-content: space-between;
+  margin-bottom: 16px;
+  padding-left: 16px;
+  padding-right: 16px;
```
Anna-Tkachenko, line 37: Same here, let's use general paddings for the card instead of such paddings for the elements

src/index.html
```diff
@@ -1,13 +1,61 @@
...
+
+        <div class="card__star-reviews">Reviews: 5</div>
+      </div>
+
+      <div class="card__price">
+        <div class="card__price-name">Price:</div>
+        <div class="card__price-number">$2,199</div>
+      </div>
+
+      <a class="card__buttom" data-qa="hover">Buy</a>
```
Anna-Tkachenko, line 57: Maybe button? (not buttom)

src/styles/card.css
```diff
@@ -0,0 +1,87 @@
...
+  text-transform: uppercase;
+  font-weight: 700;
+  font-size: 14px;
+  line-height: 40px;
+}
+
+.card__buttom:hover {
+  background-color: white;
+  color: #00acdc;
+  transition: 0.3s;
```
Anna-Tkachenko, line 86: Will be better to specify the properties, for which you use transition (for now it'll use transition for all props, and it can affect on performance in the bad way)

src/styles/star.css
```diff
@@ -0,0 +1,16 @@
+.star {
+  display: flex;
+  margin: 0 16px;
```
Anna-Tkachenko, line 3: You can't set margins to the BEM block 🥲

Anna-Tkachenko (changes requested):
Good job!
Just need to fix some moments

src/styles/card.css
```diff
@@ -0,0 +1,81 @@
...
+  color: white;
+  border: 1px solid #00acdc;
+  text-transform: uppercase;
+  font-weight: 700;
+  font-size: 14px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  line-height: 16px;
+  transition: 0.3s;
```
Anna-Tkachenko, line 75: You need to specify properties here, that can be changeable for this element (for now it uses all by default)

Anna-Tkachenko (approved):

PR #1797, opened 2022-08-18
Develop

ArturNagaycev:
- [DEMO LINK](https://ArturNagaycev.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://ArturNagaycev.github.io/layout_product-cards/report/html_report/)


src/styles/blocks/stars.css
```diff
@@ -0,0 +1,26 @@
...
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+
+  margin-right: 4px;
+
+  background-image: url(../../images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
Oleksandra-Shevchenko, line 13: you can use shorted version 

 background: url(../../images/star.svg) no-repeat center;

Oleksandra-Shevchenko (approved):
LGTM :)

src/index.html
```diff
@@ -1,13 +1,64 @@
...
+
+      <div class="card__price">
+        <div class="card__price-text">
+          Price:
+        </div>
+
+        <div class="card__price-value">
+          $2,199
+        </div>
+      </div>
```
Denchepornyuk, line 52: use more semantic tags instead of div

Denchepornyuk (changes requested):

src/index.html
```diff
@@ -1,13 +1,64 @@
...
+      <h2 class="card__title">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </h2>
+
+      <p class="card__code">
+        Product code: 195434
+      </p>
+
+      <section class="card__rating">
+        <section class="stars stars--4">
```
OleksiiBlezniuk, line 31: I don't think that section is a best choice here, section it's like a bigger part of page, probably a set of cards can be a section, but setting section for this stars block is overkill as for me, card can be article, but inside it it's okay to use divs for this purpose, just don't use divs/spans for text

```
<body>
  <header></header>
  <main>
    <section>
      <article>
        <p></p>
      </article>
    </section>
  </main>
  <footer></footer>
</body>
```

src/styles/blocks/card.css
```diff
@@ -0,0 +1,96 @@
...
+  background: #fff;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+
+.card__photo {
+  display: flex;
+  object-fit: contain;
+  width: 100%;
+  height: 174px;
```
OleksiiBlezniuk, line 14: Looks strange, this picture have it's own sizes in mockup, and indents on the bottom and on both left and right sides, so calculating of height to this image - is stranger, then just margin-bottom 🙂

src/styles/blocks/card.css
```diff
@@ -0,0 +1,96 @@
...
+
+  padding: 0 3px 40px;
+  margin: 0;
```
OleksiiBlezniuk, line 17: Do image have any default margins?

src/styles/blocks/card.css
```diff
@@ -0,0 +1,96 @@
...
+}
+
+.card__title {
+  font-size: 12px;
+  font-weight: 500;
+
+  padding-bottom: 4px;
```
OleksiiBlezniuk, line 24: As for me it's more logic to set margins to set distance between elements, I mean paddings - increases element size, so if there no need to increase it (for example clickable zone), more logic to use margins. But it's not critical, just a suggestion

src/styles/blocks/card.css
```diff
@@ -0,0 +1,96 @@
...
+  justify-content: space-between;
+  align-items: center;
+
+  padding-bottom: 16px;
+}
+
+.card__price-text {
+  color: #616070;
+
+  margin: 0;
```
OleksiiBlezniuk, line 62: Reset default margins of `p` on tag, no need to reset it in every class

src/styles/blocks/card.css
```diff
@@ -0,0 +1,96 @@
...
+  justify-content: center;
+  align-items: center;
+
+  font-weight: bold;
+  font-size: 14px;
+  text-transform: uppercase;
+  color: #fff;
+  text-decoration: none;
+
+  height: 40px;
```
OleksiiBlezniuk, line 83: As for me setting paddings is better than height hardcoding, probalby sizes of this link should be different if we'll want to change font-size

src/styles/blocks/card.css
```diff
@@ -0,0 +1,96 @@
...
+
+  background: #00acdc;
+  border-radius: 5px;
+  border: 1px solid #00acdc;
+
+  transition: 0.5s;
```
OleksiiBlezniuk, line 89: Better to define which exact properties will be animated, because if you not define it, it will be applied to every property that can be animated

OleksiiBlezniuk (changes requested):
GJ, left some comments

src/index.html
```diff
@@ -1,13 +1,64 @@
...
+
+      <section class="card__price">
+        <p class="card__price-text">
+          Price:
+        </p>
+
+        <p class="card__price-value">
+          $2,199
+        </p>
+      </section>
```
OleksiiBlezniuk, line 52: This is so small block, it's more paragraph with two spans, then section with two paragraphes

src/styles/blocks/card.css
```diff
@@ -0,0 +1,88 @@
...
+  align-items: center;
+  padding: 11px 0;
+
+  font-weight: bold;
+  font-size: 14px;
+  text-transform: uppercase;
+  color: #fff;
+  text-decoration: none;
+
+  background: #00acdc;
```
OleksiiBlezniuk, line 76:
```suggestion
  background-color: #00acdc;
```

src/styles/blocks/card.css
```diff
@@ -0,0 +1,88 @@
...
+  border: 1px solid #00acdc;
+
+  transition: background-color 0.5s,
+    color 0.5s;
+}
+
+.card__button:hover {
+  color: #00acdc;
+
+  background: #fff;
```
OleksiiBlezniuk, line 87:
```suggestion
  background-color: #fff;
```

OleksiiBlezniuk (approved):
GJ!

PR #1804, opened 2022-08-18
product cart was created

vallerii:


src/index.html
```diff
@@ -1,13 +1,76 @@
...
+    <link
+      rel="stylesheet"
+      href="./style.css"
+    >
+  </head>
+  <body>
+    <div
+      class="card"
+      data-qa="card"
+    >
```
Anna-Borodaienko, line 30:
```suggestion
    <div class="card" data-qa="card">
```
If you have only 2 properties you can put it in 1 line

src/index.html
```diff
@@ -1,13 +1,76 @@
...
+      <img 
+        class="card__img"
+        src="./images/imac.jpeg"
+        alt="iMac"
+      >
+
+      <div class="card__box">
+        <div class="card__title-1">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </div>
```
Anna-Borodaienko, line 40: Coach told me to use here h2 tag, not div
```suggestion
        <h2 class="card__title-1">
          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
        </h2>
```
From my coach review )

src/index.html
```diff
@@ -1,13 +1,76 @@
...
+
+        <div class="card__title-2">
+          Product code: 195434
+        </div>
```
Anna-Borodaienko, line 44: And here you can use tag p, not div

src/style.css
```diff
@@ -0,0 +1,124 @@
...
+  background-position: center;
+}
+  
+.stars__star:last-child {
+  margin-right: 0;
+}
+  
+.stars__star:nth-child(-n+4) {
+  background-image: url(./images/star-active.svg);
+}
```
Anna-Borodaienko, line 70: return here all this block from stars task. It has to work if you'll change this modifier in html
Reuse stars block from [Stars task](https://github.com/mate-academy/layout_stars) and keep it operational. If you change stars--4 modifier to stars--2 or other - number of selected stars should change correctly.

src/style.css
```diff
@@ -0,0 +1,124 @@
...
+
+.card__text-1 {
+  font-style: normal;
```
Anna-Borodaienko, line 73: normal is a default value, you don't need to repeat it

src/style.css
```diff
@@ -0,0 +1,124 @@
...
+  font-size: 12px;
+  color: #060b35;
+  height: 36px;
+  line-height: 18px;
+  margin-bottom: 4px;
+}
+
+.card__title-2 {
+  font-weight: 400;
+  font-size: 10px;
```
Anna-Borodaienko, line 43: You can move it to html tag and don't repeat it after in different selectors

src/style.css
```diff
@@ -0,0 +1,124 @@
...
+  align-items: center;
+  height: 38px;
+  margin-left: calc(50% - 166px / 2);
+  margin-right: calc(50% - 166px / 2);
+  bottom: 16px;
+  background: #00acdc;
+  border-radius: 5px;
+  border: 1px solid #00acdc;
+    
+  font-style: normal;
```
Anna-Borodaienko, line 112: normal is a default value, you don't need to repeat it

Anna-Borodaienko (commented):
And what about this task from checklist?
You should create separate file for each BEM block

src/index.html
vallerii: thanks, i didn`t know

src/index.html
vallerii: it isn`t rough mistake?

Anna-Borodaienko: No, but I fixed the same thing after coach review

Anna-Tkachenko: It's not a mistake, it's ok
It's recommended to put each attribute on the new line if you have 3 and more attributes
In all other cases - it's ok if you'll leave them in one line

src/index.html
```diff
@@ -1,13 +1,76 @@
...
+          <div class="stars__star"></div>
+        </div>
+
+        <div class="card__text-1">
+          Reviews: 5
+        </div>
+
+        <div class="card__text-2">
+          Price:
+        </div>
```
Anna-Tkachenko, line 60: Let's create more meaningful element names instead of text-1 and text-2 🙂

Anna-Tkachenko: And use semantic tags instead of divs

src/index.html
```diff
@@ -1,13 +1,76 @@
...
+
+        <div class="card__price" >
+          $2,199
+        </div>
```
Anna-Tkachenko, line 64: Same here, will be better to use semantic tag

src/style.css
```diff
@@ -0,0 +1,12 @@
+@import "./star.css";
+@import "./card.css";
```
Anna-Tkachenko, line 2: I don't see these files in your pull request
Try to launch `git add .`, `git commit` and `git push` one more time

Anna-Tkachenko (changes requested):

src/card.css
```diff
@@ -0,0 +1,99 @@
...
+  margin-top: 33px;
+  margin-bottom: 36px;
+}
+  
+.card__stars_box {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
+  margin-right: calc(50% - 166px / 2);
+  margin-left: calc(50% - 166px / 2);
```
Anna-Tkachenko, line 20: Let's specify correct paddings for the card class instead of adding margin-left / margin-right for each element

src/card.css
```diff
@@ -0,0 +1,99 @@
+.card {
+  width: 200px;
+  background-color: var(--color-white);
+  border: 1px solid var(--color-concrete);
+}
+  
+.card__img {
+  width: 160px;
+  height: 134px;
+  margin-left: calc(50% - 160px / 2);
```
Anna-Tkachenko, line 10: Image should be centered, don't add fixed margin-left for the image
Just add correct paddings to the main card

src/card.css
```diff
@@ -0,0 +1,99 @@
...
+  margin-left: calc(50% - 166px / 2);
+}
+  
+.card__title {
+  font-size: 12px;
+  font-weight: 500;
+  color: #060b35;
+  height: 36px;
+  line-height: 18px;
+  margin: 0;
```
Anna-Tkachenko, line 29: Why did you add margin: 0 here? if you need to rest some margins for some elements, use reset.css file

src/star.css
```diff
@@ -0,0 +1,25 @@
+.stars--4 {
+  height: 16px;
+  display: flex;
+}
```
Anna-Tkachenko, line 4: These styles are redundant, delete them

Anna-Tkachenko (changes requested):

src/card.css
vallerii: i have to use margin: 0 calc(50% - 166px / 2); or dont use this formuls?

src/star.css
vallerii: i can delete only "height", if i`ll delete "display" stars are breaking

Anna-Tkachenko: @vallerii so move this property to some other selector. If I'll change modifier to stars--3, everything will be broken

src/reset.css
```diff
@@ -0,0 +1,11 @@
+body {
+  margin: 0;
+  width: 200px;
```
Anna-Tkachenko, line 3: you shouldn't add any width to the body, delete this property from here
you should just specify the width for the product card

src/reset.css
```diff
@@ -0,0 +1,11 @@
...
+}
+
+.card__title,
+.card__price,
+.card__price_value,
+.card__reviews {
```
Anna-Tkachenko, line 9: use tag selectors here (p, ul, for example, or any else, that you need)
don't use elements classes in reset.css file

src/card.css
```diff
@@ -0,0 +1,89 @@
+.card {
+  box-sizing: border-box;
+  width: 200px;
+  background-color: #fff;
+  border: 1px solid #f3f3f3;
+  padding: 32px 16px 16px;
```
Anna-Tkachenko, line 6: I don't see paddings on the demo link, did you run deploy command one more time after you changes? If no, please run 'npm run deploy' and check that all changes will be applied on the demo link

src/card.css
```diff
@@ -0,0 +1,89 @@
...
+  line-height: 18px;
+}
+  
+.card__code {
+  font-size: 10px;
+  line-height: 14px;
+  color: #616070;
+  height: 14px;
+  margin-bottom: 16px;
+  margin-top: 4px;
```
Anna-Tkachenko, line 35: check point 10 in checklist
you shouldn't add both margin-top and margin-bottom to the elements
the logic should be the next: you add correct paddings to the whole card, then add all needed elements one by one, and specify correct margin-bottom for each element
if you'll have any questions about implementation, ask please in main channel for questions ❤️

Anna-Tkachenko (changes requested):

src/card.css
vallerii: yheah, i did it, but ill do it one more time

src/reset.css
```diff
@@ -0,0 +1,11 @@
...
+  margin: 0;
+}
+
+h2 {
+  margin: 0;
+}
+
+p {
+  margin: 0;
+}
```
Anna-Tkachenko, line 11: you can combine selectors using coma
```suggestion
body, h2, p {
  margin: 0;
}
```

Anna-Tkachenko (approved):
Good job!

PR #1806, opened 2022-08-18
 layout_product-cards

xiimiik:
- [DEMO LINK](https://xiimiik.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://xiimiik.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -1,0 +1,3 @@
+@import "./styles/reset.css";
+@import "./styles/stars.css";
+@import "./styles/card.css";
--- a/src/index.css
+++ b/src/index.css
@@ -1,3 +1,4 @@
 @import "./styles/reset.css";
 @import "./styles/stars.css";
 @import "./styles/card.css";
+@import "./styles/variables.css";
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,42 @@
-<!doctype html>
-<html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <title>Product cards</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Product cards</h1>
-</body>
+<!DOCTYPE html>
+<html lang="en" class="page">
+  <head>
+    <meta charset="UTF-8">
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
+    <title>Product cards</title>
+    <link rel="stylesheet" href="index.css">
+  </head>
+  <body>
+    <div class="card">
+      <img
+        src="./images/imac.jpeg"
+        alt="Imac image"
+        class="card__image"
+      >
+
+      <span> APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A) </span>
+
+      <span> Product code: 195434 </span>
+
+      <div class="card__reviews">
+        <div class="stars stars--4">
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+        </div>
+
+        <span>Reviews: 5</span>
+      </div>
+
+      <div>
+        <span>Price:</span>
+        <span>$2,199</span>
+      </div>
+    </div>
+  </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -18,19 +18,19 @@
         class="card__image"
       >
 
-      <span class="card__title">
+      <h3 class="card__title">
         APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
-      </span>
+      </h3>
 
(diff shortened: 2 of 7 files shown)


src/index.css
```diff
@@ -0,0 +1,4 @@
+@import "./styles/reset.css";
+@import "./styles/stars.css";
+@import "./styles/card.css";
+@import "./styles/variables.css";
```
anton-iskryk, line 4: Just a tip - you can add comments to describe the files which you imported. For example, before first import add comment /* RESET SETTINGS */  etc. If you don't agree - you don't have to add comments like this.

src/index.html
```diff
@@ -1,13 +1,64 @@
...
+        </p>
+      </div>
+
+      <a
+        href="#"
+        class="card__button"
+        data-qa="hover"
+      >
+        <p class="card__button-text">buy</p>
+      </a>
```
anton-iskryk, line 61: I think there's no need to add a separate <p> element for the text - you can just add the word 'buy' to anchor element.

src/styles/card.css
```diff
@@ -0,0 +1,93 @@
...
+.card__rating {
+  align-items: center;
+  display: flex;
+  width: 100%;
+  justify-content: space-between;
+}
+
+.card__price {
+  margin-bottom: 16px;
+}
```
anton-iskryk, line 57: It's quite confusing when you use twice the same elements even when styles are repeated. I think it's better to write .card__price and .card__rating separately.

src/styles/reset.css
```diff
@@ -0,0 +1,9 @@
+body,
+h3,
+p {
+  margin: 0;
+}
+
+.html {
+  font-family: Roboto, sans-serif;
+}
```
anton-iskryk, line 9: Do we need to add this to the 'reset.css' file? I think it should be placed in the card.css.

anton-iskryk (changes requested):
Just a couple of comments on the code. Write if you don't agree.

src/index.html
xiimiik: Forgot to delete))) just wanted to style the text separately

src/styles/card.css
xiimiik: Yes, i thinked about that, but I didn't want to repeat the same code. That's why I wrote like this, I don't even know how to do it better, so "Маємо те, те що маємо" )))

src/index.css
xiimiik: Great tip, thank you, I'll remember it

anton-iskryk (approved):

src/index.html
```diff
@@ -1,13 +1,64 @@
...
+      </h3>
+
+      <p class="card__product-code">
+        Product code: 195434
+      </p>
+
+      <div class="card__rating" >
+        <div
+        class="card__stars stars stars--4"
+        >
```
OleksiiBlezniuk, line 32:
```suggestion
        <div class="card__stars stars stars--4">
```

OleksiiBlezniuk: You can place it on one line, it's only one attribute here and that line is looks shorter than 80 symbols

src/styles/card.css
```diff
@@ -0,0 +1,93 @@
...
+  font-size: 12px;
+  line-height: 18px;
+  color: var(--color-accent);
+}
+
+.card__product-code {
+  width: 100%;
+  margin-bottom: 16px;
+
+  font-weight: 400;
```
OleksiiBlezniuk, line 32: It's a default value, no need to duplicate

src/styles/card.css
```diff
@@ -0,0 +1,93 @@
...
+  font-size: 10px;
+  color: var(--color-secondary);
+}
+
+.card__rating {
+  margin-bottom: 24px;
+}
+
+.card__reviews {
+  font-weight: 400;
```
OleksiiBlezniuk, line 42: It's a default value, no need to duplicate

src/styles/card.css
```diff
@@ -0,0 +1,93 @@
...
+  width: 100%;
+  justify-content: space-between;
+}
+
+.card__price {
+  margin-bottom: 16px;
+}
+
+.card__price-text {
+  font-weight: 400;
```
OleksiiBlezniuk, line 60: It's a default value, no need to duplicate

src/styles/card.css
```diff
@@ -0,0 +1,93 @@
...
+  border-radius: 5px;
+
+  text-decoration: none;
+  font-size: 14px;
+  font-weight: 700;
+  text-transform: uppercase;
+  color: var(--color-white);
+  background-color: var(--color-blue);
+  border: 1px solid var(--color-blue);
+  transition: 0.5s;
```
OleksiiBlezniuk, line 87: Better to set exact property that you want to animate, because if not, this transition will applied to every property

src/styles/card.css
```diff
@@ -0,0 +1,93 @@
+.card {
+  box-sizing: border-box;
```
OleksiiBlezniuk, line 2: Set it to * selector, you've applied this rule to more than one element

OleksiiBlezniuk (changes requested):
GJ! Left some comments to fix

src/index.html
xiimiik: I also wanted one line, but lint gave me an error if I wrote in one line

OleksiiBlezniuk (approved):
👍

PR #1813, opened 2022-08-19
Develop

DenysKleimenov:
- [DEMO LINK](https://DenysKleimenov.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://DenysKleimenov.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,54 @@
 <!doctype html>
-<html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <title>Product cards</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Product cards</h1>
-</body>
+<html lang="en" class="html">
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no,
+          initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <title>Product cards</title>
+    <link rel="stylesheet" href="./styles/index.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+  </head>
+  <body>
+    <div class="card" data-qa="card">
+      <div class="card__info">
+        <img
+          src="./images/imac.jpeg"
+          alt="imac photo"
+          class="card__info-logo"
+        >
+        <h2 class="card__info-title">APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</h2>
+        <p class="card__info-code">Product code: 195434</p>
+
+        <div class="card__review">
+          <div class="stars stars--4">
+            <span class="stars__star"></span>
+            <span class="stars__star"></span>
+            <span class="stars__star"></span>
+            <span class="stars__star"></span>
+            <span class="stars__star"></span>
+          </div>
+
+          <p class="card__review-mark">
+            Reviews: 5
+          </p>
+        </div>
+
+        <div class="card__price">
+          <p class="card__price-text">Price:</p>
+          <p class="card__price-amount">$2,199</p>
+        </div>
+
+        <a
+          href="#buy"
+          class="card__button"
+          data-qa="hover"
+        >
+          Buy
+        </a>
+      </div>
+    </div>
+  </body>
 </html>
--- a/src/styles/card.css
+++ b/src/styles/card.css
@@ -1,0 +1,103 @@
+/*  CARD BLOCK:  */
+
+.card {
+  padding: 32px 0 16px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  width: 198px;
+}
+
+.card__info {
+  display: flex;
+  flex-direction: column;
+  margin: 0 16px;
+}
+
+.card__info-logo {
+  width: 160px;
+  height: 134px;
+  margin: 0 auto 40px;
+}
+
+.card__info-title {
+  color: #060b35;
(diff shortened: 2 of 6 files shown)


src/index.html
```diff
@@ -1,13 +1,54 @@
...
+          initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <title>Product cards</title>
+    <link rel="stylesheet" href="./styles/index.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+  </head>
+  <body>
+    <div class="card" data-qa="card">
+      <div class="card__info">
```
grafit-off, line 16: You can do it without a `card__info` div, but it is not critical.

src/styles/card.css
```diff
@@ -0,0 +1,105 @@
+/*  CARD BLOCK:  */
+
+.card {
+  padding: 32px 0 16px;
+  border: 1px solid #f3f3f3;
```
grafit-off, line 5: You can use variable for colors.

src/styles/card.css
```diff
@@ -0,0 +1,105 @@
...
+  font-weight: 700;
+  font-size: 14px;
+  font-family: inherit;
+  line-height: 16px;
+  text-transform: uppercase;
+
+  border: 1px solid #00acdc;
+  border-radius: 5px;
+
+  transition: all 0.5s ease-out;
```
grafit-off, line 99: Better to define which exact properties will be animated, because if you not define it, it will be applied to every property that can be animated.

grafit-off (approved):
Good job. I'm left some comments below 🔥

src/styles/card.css
OleksiiBlezniuk: Agree with @grafit-off

src/index.html
OleksiiBlezniuk: Agree with @grafit-off , no need to use any extra containers here

src/index.html
```diff
@@ -1,13 +1,54 @@
...
+        <img
+          src="./images/imac.jpeg"
+          alt="imac photo"
+          class="card__info-logo"
+        >
+        <h2 class="card__info-title">APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</h2>
```
OleksiiBlezniuk, line 22: This line looks too long

OleksiiBlezniuk, line 22:
```suggestion
        <h2 class="card__info-title">
          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
        </h2>
```

src/styles/card.css
```diff
@@ -0,0 +1,105 @@
+/*  CARD BLOCK:  */
+
+.card {
+  padding: 32px 0 16px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  width: 198px;
```
OleksiiBlezniuk, line 7: Better to set `box-sizing: border-box`, then calculate needed width to fit borders

src/styles/card.css
```diff
@@ -0,0 +1,105 @@
...
+}
+
+.card__info {
+  display: flex;
+  flex-direction: column;
+  margin: 0 16px;
```
OleksiiBlezniuk, line 13: Strange solution. It's a fixed distance between `card` borders and it's content, so padding for card looks like a good solution here, you also set vertical paddings to container, so why not set horizontal too 🙂

src/styles/card.css
```diff
@@ -0,0 +1,105 @@
...
+}
+
+.card__info-title {
+  color: var(--dark-color);
+  font-family: inherit;
+  font-weight: 500;
+  font-size: 12px;
+
+  line-height: 18px;
+  margin: 0 0 4px;
```
OleksiiBlezniuk, line 31: As for me, it's better to reset default margins somewhere in `reset` file and set margins only for that sides that you need (it's more readable)

OleksiiBlezniuk: But it's not critical here

src/styles/card.css
```diff
@@ -0,0 +1,105 @@
...
+}
+
+.card__info-code {
+  font-family: inherit;
+  font-size: 10px;
+  line-height: 14px;
+  color: var(--grey-color);
+  margin: 0 0 16px;
```
OleksiiBlezniuk, line 39: You've reset margins for `p`, so lets set only margin-bottom

src/styles/card.css
```diff
@@ -0,0 +1,105 @@
...
+  border-radius: 5px;
+  width: 198px;
+}
+
+.card__info {
+  display: flex;
+  flex-direction: column;
+  margin: 0 16px;
+  --dark-color: #060b35;
+  --grey-color: #616070;
```
OleksiiBlezniuk, line 15: Better to set css variables to `:root` to make it possible to use it not only in `card__info` but in any element on the page

OleksiiBlezniuk (changes requested):
GJ, left some comments

src/index.html
```diff
@@ -1,13 +1,54 @@
...
+        </p>
+      </div>
+
+      <div class="card__price">
+        <p class="card__price-text">Price:</p>
+        <p class="card__price-amount">$2,199</p>
+      </div>
+
+      <a
+        href="#buy"
```
OleksiiBlezniuk, line 46:
```suggestion
        href="#"
```

OleksiiBlezniuk: There is no `buy` section on your page, so `#` is enough

src/styles/card.css
```diff
@@ -0,0 +1,101 @@
...
+.card__price-amount {
+  font-family: inherit;
+  font-weight: 700;
+  font-size: 16px;
+  line-height: 18px;
+  color: var(--dark-color);
+}
+
+.card__button {
+  height: 38px;
```
OleksiiBlezniuk, line 76: Screenshot 2022-08-22 at 17 10 11

OleksiiBlezniuk (approved):
Would be better to use more meaningful messages to commit, it should shortly describe you changes. I can't understand what was changed from `Problems were fixed`

PR #1868, opened 2022-08-31
task

NVlasenko:
- [DEMO LINK](https://NVlasenko.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://NVlasenko.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,142 @@
+p {
+  margin: 0;
+}
+
+h1 {
+  margin: 0;
+}
+
+h2 {
+  margin: 0;
+}
+
+body {
+  margin: 0 auto;
+}
+
+.block {
+  box-sizing: border-box;
+  font-family: "Roboto", sans-serif;
+  display: flex;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,142 @@
...
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+  border-radius: 0.5px;
+
+}
+
+.stars__star--active :nth-child(-n + 4) {
+  background-image: url(../src/images/star-active.svg) ;
+}
```
ArturErmolenko, line 78: reuse all styles from Stars tasks,it should change correctly active stars if we change modifier

ArturErmolenko (changes requested):
1.You should create separate file for each BEM block
2.
Screenshot 2022-09-01 at 13 58 15

ArturErmolenko (approved):
Well done 🔥

PR #1872, opened 2022-09-02
card was created

katya-suvorova:


src/card.css
```diff
@@ -0,0 +1,113 @@
...
+}
+
+.card__product-name {
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 18px;
+  color: #060b35;
+  margin: 0 16px 4px;
+}
```
Bohdan-Maksymiuk, line 26: you could set horizontal padding for the whole card instead of setting margins for every element

src/card.css
```diff
@@ -0,0 +1,113 @@
...
+
+.card__reviews {
+  color: #060b35;
+  font-style: normal;
+  font-weight: 400;
+  font-size: 10px;
+  line-height: 14px;
+  width: 53px;
+  height: 16px;
```
Bohdan-Maksymiuk, line 51: do not hardcode size of element

src/card.css
```diff
@@ -0,0 +1,113 @@
...
+}
+
+.card__price-container {
+  display: flex;
+  justify-content: space-between;
+  width: 166px;
```
Bohdan-Maksymiuk, line 57: same

src/card.css
```diff
@@ -0,0 +1,113 @@
...
+  height: 18px;
+}
+
+.card__price-number {
+  font-style: normal;
+  font-weight: 700;
+  font-size: 16px;
+  line-height: 18px;
+  width: 50px;
+  height: 18px;
```
Bohdan-Maksymiuk, line 77: do not hardcode size of element

src/index.html
```diff
@@ -1,13 +1,60 @@
...
+    <div class="card" data-qa="card">
+      <img 
+        src="./images/imac.jpeg" 
+        alt="card-image" 
+        class="card__image"
+      >
+
+      <p class="card__product-name">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </p>
```
Bohdan-Maksymiuk, line 21: <h2> will be semantically more correct

Bohdan-Maksymiuk (changes requested):

src/card.css
katya-suvorova: updated

src/card.css
katya-suvorova: updated

src/card.css
katya-suvorova: updated

src/card.css
katya-suvorova: updated

src/index.html
katya-suvorova: updated

Virass (approved):

PR #1958, opened 2022-10-02
solution

mitattt:
- [DEMO LINK](https://mitattt.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://mitattt.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,65 @@
 <!doctype html>
-<html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <title>Product cards</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Product cards</h1>
-</body>
+<html lang="en" class="page">
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width,
+          user-scalable=no,
+          initial-scale=1.0,
+          maximum-scale=1.0,
+          minimum-scale=1.0"
+    >
+    <title>Product cards</title>
+    <link rel="stylesheet" href="./index.css">
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="anonymous"
+    >
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+  </head>
+  <body>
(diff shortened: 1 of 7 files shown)


src/index.html
```diff
@@ -1,13 +1,65 @@
...
+    >
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+  </head>
+  <body>
+    <section class="card" data-qa="card">
+      <img
+        src="./images/imac.jpeg"
+        alt="imac"
+        class="card__image"
+      >
```
masha-girya, line 27: you made blank lines everywhere between sibling elements except there. I think you should add it theree too

src/styles/card.css
```diff
@@ -0,0 +1,94 @@
...
+  padding: 32px 16px 0;
+  border-radius: 5px;
+  border: 1px solid var(--light-white);
+  width: var(--card-width);
+  background-color: var(--white);
+}
+
+.card__image {
+  width: 160px;
+  height: 134px;
```
masha-girya, line 14: CHECKLST 8. [STYLES] - Don't hardcode exact height size. Add necessary paddings according to mockup and let content dictate the container size.

src/styles/card.css
```diff
@@ -0,0 +1,94 @@
...
+  text-transform: uppercase;
+  outline: none;
+
+  height: 40px;
+  line-height: 16px;
+  font-size: 14px;
+  font-weight: 700;
+  margin-bottom: 16px;
+  border-radius: 5px;
+  border: 1px transparent;
```
masha-girya, line 84: as I understand, we should make border 1px solid transparent and then in :hover change it like: border-color: value;

masha-girya (changes requested):
Greate job, i added some comments :) I liked that you used variables !!

src/styles/card.css
mitattt: > CHECKLST 8. [STYLES] - Don't hardcode exact height size. Add necessary paddings according to mockup and let content dictate the container size.

I did this to set the size of the picture but not the block

src/index.html
mitattt: ok i forgot about this

src/styles/card.css
mitattt: it does not work in this way

mitattt: i dont agree with you

masha-girya: border: 1px solid transparent;

masha-girya (changes requested):
only border left (checklist 9)

masha-girya (approved):
Cool!!

PR #1969, opened 2022-10-03
Develop

Stepico:
- [DEMO LINK](https://Stepico.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://Stepico.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/styles/card.css
+++ b/src/styles/card.css
@@ -8,11 +8,12 @@
 
 .card__img {
   width: 100%;
+  height: 134px;
   padding: 0 3px;
 }
 
 .card__name {
-  margin-top: 37px;
+  margin-top: 36px;
 
   font-weight: 500;
   font-size: 12px;


src/index.html
```diff
@@ -1,13 +1,54 @@
...
+    <link rel="stylesheet" href="./style.css">
+  </head>
+  <body class="home-page__content">
+    <div class="card" data-qa="card">
+      <img
+        src="./images/imac.jpeg"
+        alt="IMac"
+        class="card__img"
+      >
+      <div class="card__name">
```
MaxFedak, line 17: due to checklist we have to use semantics (h2, p) with classes for text
here and further

src/index.html
```diff
@@ -1,13 +1,54 @@
...
+      </div>
+      <div class="card__price">
+        <div class="card__price-tag">
+          <span>Price:</span>
+        </div>
+        <div class="card__price-money">
+          <span>$2,199</span>
+        </div>
+      </div>
+      <button class="card__btn">
```
MaxFedak, line 43: i guess it would be better to use just a link and style it
in that way we will get rid of some unnecessary nesting

src/styles/card.css
```diff
@@ -0,0 +1,90 @@
+.card {
+  width: 200px;
+  padding: 16px;
+  padding-top: 32px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+
+.card__img {
```
MaxFedak, line 9: not sure if that's right, but if you add a height from scratch, it will be a perfect match for test

src/index.html
```diff
@@ -1,13 +1,54 @@
...
+        alt="IMac"
+        class="card__img"
+      >
+      <div class="card__name">
+        <span>APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</span>
+      </div>
+      <div class="card__code">
+        <span>Product code: 195434</span>
+      </div>
+      <div class="card__rate">
```
MaxFedak, line 23: i guess if u put stars block and reviews element in flex element, you will be able to position them more accurate due to mockup

src/styles/card.css
```diff
@@ -0,0 +1,90 @@
...
+}
+
+.card__price-money {
+  font-weight: 700;
+  font-size: 16px;
+  line-height: 18px;
+  color: #060b35;
+}
+
+.card__btn {
```
MaxFedak, line 63: smth is wrong with font. It doesnt inherit

src/styles/card.css
```diff
@@ -0,0 +1,90 @@
...
+  width: 100%;
+  padding: 0 3px;
+}
+
+.card__name {
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 18px;
+  color: #060b35;
+  margin-top: 37px;
```
MaxFedak, line 19: just curios why its 37px?

Stepico: I was trying to find the perfect match so that all tests could be run successfully :)

src/index.html
Stepico: Yeah, but they have their own default values for padding and margin, so I've just decided to use span instead

src/index.html
mykola-koval: agree

src/styles/card.css
mykola-koval: yes, you can add height for image according to the design

mykola-koval (changes requested):
seems like some styles are wrong, check the test and fix them 🚀 
 - reviews text
 - buy button text
 - image height

mykola-koval (approved):

PR #1990, opened 2022-10-04
Product-card was added

a-kulaev:
https://a-kulaev.github.io/layout_product-cards/
https://a-kulaev.github.io/layout_product-cards/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,66 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <title>Product cards</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Product cards</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -1,13 +1,66 @@
...
+      class="product-card"
+      data-qa="card"
+    >
+      <img
+        src="./images/imac.jpeg"
+        alt="iMac"
+        class="product-card__pic product-card__pic--position"
+      >
+      </img>
+      <div class="product-card__header product-card__header--position">
```
liliia-livandovskaya, line 28:
```suggestion
      </img>
      
      <div class="product-card__header product-card__header--position">
```
pay attention to the code style, fix it all over the page

src/styles/blocks/product-card.css
```diff
@@ -0,0 +1,101 @@
+.product-card {
+  display: flex;
+  flex-direction: column;
+  width: 200px;
+  height: 406px;
```
liliia-livandovskaya, line 5: checklist p.8 [STYLES] - Don't hardcode exact height size. Add necessary paddings according to mockup and let content dictate the container size

liliia-livandovskaya (changes requested):
GJ  Pay attention to the comments

src/styles/blocks/product-card.css
a-kulaev: @liliia-livandovskaya

1. I already set paddings. Do I need to change "height" to "min-height", so in case if I get additional elements, card could get higher? Did I get it correct?

2. Can I leave fixed width, or I need to figure out another way to keep my card size? When I remove width, my card grow to full body width.

liliia-livandovskaya (approved):

PR #2010, opened 2022-10-09
productCardCreated

AllaNebozhenko:
[DEMO LINK](https://AllaNebozhenko.github.io/layout_product-cards/)
[TEST REPORT LINK](https://AllaNebozhenko.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/styles/card.css
+++ b/src/styles/card.css
@@ -1,0 +1,74 @@
+
+.card {
+  flex-direction: column;
+  width: 200px;
+  padding: 16px;
+  background: #fff;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  font-style: normal;
+  font-family: Roboto, sans-serif;
+}
+
+.card__image {
+  width: 160px;
+  height: 134px;
+  margin: 16px 3px;
+}
+
+.card__name {
+  margin-top: 24px;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 18px;
+  color: #060b35;
+}
+
+.card__code {
+  font-weight: 400;
+  font-size: 10px;
+  line-height: 14px;
+  color: #616070;
+  margin-top: 4px;
+}
+
+.card__price {
+  justify-content: space-between;
+  align-items: center;
+  line-height: 18px;
+  margin-top: 24px;
+}
+
+.card__price-name {
+  font-weight: 400;
+  font-size: 12px;
+  color: #616070;
+}
+
+.card__price-amount {
+  font-weight: 700;
+  font-size: 16px;
+  color: #060b35;
+}
+
+.card__link {
+  text-decoration: none;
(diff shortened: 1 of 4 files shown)


AllaNebozhenko:
Доброго дня! Не можу більше зрозуміти, що в шрифтах змінити, щоб повністю зійшлись тести. Допоможіть

src/styles/card.css
```diff
@@ -0,0 +1,74 @@
+
```
marina-tilniak, line 1:
```suggestion
```

src/styles/stars.css
```diff
@@ -0,0 +1,36 @@
+
+.card__stars {
```
marina-tilniak, line 2:
```suggestion
.card__stars {
```

src/styles/stars.css
```diff
@@ -0,0 +1,36 @@
...
+.stars--5 :nth-child(-n+5) {
+  background: url(/images/star-active.svg) no-repeat center;
+}
+
+.card__review {
+  font-weight: 400;
+  font-size: 10px;
+  line-height: 14px;
+  margin-top: 16px;
+}
```
marina-tilniak, line 36: it's a part of card

src/styles/card.css
```diff
@@ -0,0 +1,74 @@
...
+.card__price-amount {
+  font-weight: 700;
+  font-size: 16px;
+  color: #060b35;
+}
+
+.card__link {
+  text-decoration: none;
+  width: 166px;
+  height: 40px;
```
marina-tilniak, line 57: let's  use padding instead of height

marina-tilniak (changes requested):
what is this?)
image
it's wrong font, check it
image

src/styles/stars.css
```diff
@@ -0,0 +1,28 @@
+.card__stars {
```
AlisaVlasova, line 1:
```suggestion
.stars {
```

AlisaVlasova (changes requested):
still here
image

src/styles/stars.css
AllaNebozhenko: Hi ! ok, help me. First I add Roboto via @import, than i changed it like there (https://www.codegrepper.com/code-examples/css/css+roboto+font)  in <head> and html. How to do it correct? thanks

nazarmatsevych: It's better you use link if you are loading it from CDN or google fonts. If the font is in your local directory then loading it using import or link will not have too much significant difference. But if you are loading it using a third party then it's always better to use link .

If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review 🙂

nazarmatsevych (approved):
Good work!

PR #2053, opened 2022-10-25
Develop

diamat1997:
- [DEMO LINK](https://diamat1997.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://diamat1997.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,146 @@
+@import url(https://fonts.googleapis.com/css2?family=Roboto:wght@100;300;400;500;700;900&display=swap);
+
+body {
+  font-family: Roboto, sans-serif;
+  padding: 0;
+  margin: 0;
+}
+
+.card {
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-content: center;
+  align-items: center;
+  width: 200px;
+  border: 1px #f3f3f3 solid;
+}
+
+.card__img {
+  display: block;
+  margin: 0 auto;
+  width: 160px;
+  height: 134px;
+}
+
+.card__img_link {
+  margin: 32px 0 40px;
+}
+
+.card__name_link {
+  text-decoration: none;
+  color: #060b35;
+  width: 166px;
+  margin-bottom: 4px;
+}
+
+.card__name_link_text {
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 18px;
+  margin: 0;
+}
+
+.card__product_code {
+  font-size: 10px;
+  line-height: 14px;
+  color: #616070;
+  width: 166px;
+  margin-bottom: 16px;
+  margin-top: 0;
+}
+
+.card__reviews {
+  display: flex;
+  justify-content: space-between;
+  align-items: flex-end;
+  align-content: flex-end;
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,146 @@
+@import url(https://fonts.googleapis.com/css2?family=Roboto:wght@100;300;400;500;700;900&display=swap);
+
+body {
+  font-family: Roboto, sans-serif;
+  padding: 0;
```
SanyaBratashchuk, line 5: body doesn't have default paddings

src/style.css
```diff
@@ -0,0 +1,146 @@
...
+  margin-right: 0;
+}
+
+.stars--1 :nth-child(1),
+.stars--2 :nth-child(-n + 2),
+.stars--3 :nth-child(-n + 3),
+.stars--4 :nth-child(-n + 4),
+.stars--5 :nth-child(-n + 5) {
+  background-image: url(/images/star-active.svg);
+}
```
SanyaBratashchuk, line 146: move each BEM block to the seperate file

SanyaBratashchuk (changes requested):

src/style.css
diamat1997: Снимок
I did as you said, transferred the code to another file, the styles are pulled up, but now the tests do not pass. If I return the same code, the tests pass.

src/style.css
```diff
@@ -0,0 +1,121 @@
...
+  line-height: 18px;
+  color: #060b35;
+  margin: 0;
+}
+
+.card__buy_button {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: 164px;
```
AlisaVlasova, line 100: do not hardcode width, use card paddings to set right spacing, like this
image

src/style.css
```diff
@@ -0,0 +1,121 @@
...
+  height: 38px;
```
AlisaVlasova, line 101:
```suggestion
  height: 40px;
```
add * {box-sizing: border-box;}

src/style.css
```diff
@@ -0,0 +1,121 @@
...
+  margin-bottom: 16px;
+  text-transform: uppercase;
+  font-style: normal;
```
AlisaVlasova, line 104: this is default

src/style.css
```diff
@@ -0,0 +1,121 @@
+@import url(https://fonts.googleapis.com/css2?family=Roboto:wght@100;300;400;500;700;900&display=swap);
+
+body {
+  font-family: Roboto, sans-serif;
+}
+
+.card {
+  box-sizing: border-box;
```
AlisaVlasova, line 8: move it to *

src/style.css
```diff
@@ -0,0 +1,121 @@
+@import url(https://fonts.googleapis.com/css2?family=Roboto:wght@100;300;400;500;700;900&display=swap);
+
+body {
+  font-family: Roboto, sans-serif;
```
AlisaVlasova, line 4: move it to html selector, add fallback font

AlisaVlasova (changes requested):
check font-weight or color to fix tests
image

PR #2095, opened 2022-11-02
Develop

pushkarskiyrodion:
[DEMO LINK](https://pushkarskiyrodion.github.io/layout_product-cards/)
[TEST REPORT LINK](https://pushkarskiyrodion.github.io/layout_product-cards/report/html_report/)


DarkMistyRoom (changes requested):
- When tests can't find the reference file, delete backstop_data folder and run them again
- It is better to make spaces between elements with margins. and you don't need to set margin or padding to each element if you just set padding for the container. like here
image

src/index.html
```diff
@@ -1,13 +1,51 @@
...
+  </head>
+  <body>
+    <article class="card" data-qa="card">
+      <img
+        class="card__product-image"
+        src="./images/imac.jpeg"
+        alt="imac"
+      >
+
+      <div class="card__description">
```
SanyaBratashchuk, line 20: why do you need this `div`

SanyaBratashchuk (approved):

PR #2115, opened 2022-11-05
product card was created

oirevlis1:
https://oirevlis1.github.io/layout_product-cards/
https://oirevlis1.github.io/layout_product-cards/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,118 @@
+.page__content {
+  margin: 0;
+  font-family: Roboto, Arial, sans-serif;
+}
+
+.card {
+  width: 200px;
+  height: 408px;
+  box-sizing: border-box;
+  background-color: #fff;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,118 @@
...
+
+.card__img {
+  width: 160px;
+  height: 134px;
+  margin-top: 32px;
+  margin-left: 19px;
+}
+
+.card__description {
+  box-sizing: border-box;
```
TarasHoliuk, line 23: move it to `*` selector (should be applied to all elements)

TarasHoliuk (changes requested):

src/style.css
oirevlis1: '+'

nazarmatsevych (approved):
Good work!

PR #2227, opened 2022-11-30
Html and CSS files was added

Vasyl-Pavlenko:
- [DEMO LINK](https://vasyl-pavlenko.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://vasyl-pavlenko.github.io/layout_product-cards/report/html_report/)

NPM TEST з невідомих мені причин не спрацював. Порадили відправляти так.

### DIFF
--- a/src/styles/card.css
+++ b/src/styles/card.css
@@ -1,0 +1,96 @@
+.card {
+  font-family: Roboto, sans-serif;
+  display: flex;
+  width: 200px;
+  flex-direction: column;
+  background: #fff;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  padding: 16px;
+}
+
+.card__image {
+  width: 160px;
+  height: 134px;
+  margin: 16px 3px 0;
+}
+
+.card__title {
+  margin: 40px auto 0;
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 18px;
+  color: var(--mainTextColor);
+}
+
+.card__subtitle {
+  margin: 4px 0 0;
+  font-style: normal;
+  font-weight: 400;
+  font-size: 10px;
+  line-height: 14px;
+  color: var(--secondaryTextColor);
+}
+
+.card__raiting {
+  display: flex;
+  justify-content: space-between;
+  margin: 16px 0 0;
+}
+
+.card__reviews {
+  font-style: normal;
+  font-weight: 400;
+  font-size: 10px;
+  line-height: 14px;
+  color: var(--mainTextColor);
+}
+
+.card__prices {
+  margin: 24px 0 0;
+  display: flex;
+  justify-content: space-between;
+}
+
+.card__price {
+  font-style: normal;
+  font-weight: 400;
+  font-size: 12px;
+  line-height: 18px;
+  color: var(--secondaryTextColor);
+}
+
+.card__value {
+  font-style: normal;
+  font-weight: 700;
+  font-size: 16px;
+  line-height: 18px;
+  color: var(--mainTextColor);
+}
+
+.card__link {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  text-align: center;
+  text-transform: uppercase;
+  margin: 16px 0 0;
+  width: 166px;
+  height: 40px;
+  background: #00acdc;
+  border-radius: 5px;
+  color: #fff;
+  font-style: normal;
+  font-weight: 700;
+  font-size: 14px;
+  line-height: 16px;
+  text-decoration: none;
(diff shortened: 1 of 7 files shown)


src/styles/card.css
```diff
@@ -0,0 +1,96 @@
...
+}
+
+.card__raiting {
+  display: flex;
+  justify-content: space-between;
+  margin: 16px 0 0;
+}
+
+.card__reviews {
+  font-style: normal;
```
OlEzhA0, line 43: why do we need it here?

src/styles/card.css
```diff
@@ -0,0 +1,96 @@
...
+}
+
+.card__prices {
+  margin: 24px 0 0;
+  display: flex;
+  justify-content: space-between;
+}
+
+.card__price {
+  font-style: normal;
```
OlEzhA0, line 57: the same, check this fonts styles and move them to `body` for example

src/styles/card.css
```diff
@@ -0,0 +1,96 @@
...
+  font-size: 16px;
+  line-height: 18px;
+  color: var(--mainTextColor);
+}
+
+.card__link {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  text-align: center;
```
OlEzhA0, line 76: it's redundant if we center text with flex

src/styles/card.css
```diff
@@ -0,0 +1,96 @@
...
+  text-transform: uppercase;
+  margin: 16px 0 0;
+  width: 166px;
```
OlEzhA0, line 79: we don't need define width here as well

src/styles/card.css
```diff
@@ -0,0 +1,96 @@
...
+  color: #fff;
+  font-style: normal;
+  font-weight: 700;
+  font-size: 14px;
+  line-height: 16px;
+  text-decoration: none;
+  border: 1px solid transparent;
+}
+
+.card__link:hover {
```
OlEzhA0, line 92: add transition effect on hover

OlEzhA0 (changes requested):
Hey! Good job! check my comments below

src/styles/stars.css
```diff
@@ -0,0 +1,21 @@
...
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(../images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
+  border-radius: 0.5px;
```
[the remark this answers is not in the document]
Vasyl-Pavlenko: In the task was told: "Reuse stars block from [Stars task]" (https://github.com/mate-academy/layout_stars)'.
Border-radius was added to that block.  https://www.figma.com/file/EIBkG1dy1jnK88YPO34Qir/Moyo-%2F-Catalog-(updated)?node-id=11325%3A2961&t=MyQNUIAgRCGo4BEC-0

src/styles/card.css
Vasyl-Pavlenko: Sorry, but I don't understand.
In task was told: "follow styles from the mock".
Width was taken from there. 
This is not problem for me to delete width and add margin auto.
I think i need to visit Q&A meeting.

src/styles/card.css
Vasyl-Pavlenko: I have question.
Where i can take value for transition?
Maybe i missed it.
Thank you.

src/styles/stars.css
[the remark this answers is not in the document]
Viktor-Kost: Perhaps, first time mentor didn't notice that you had used border-radius. 
I checked a figma maket. And it doesn't make sense to add border-radius here as [OlEzhA0](https://github.com/OlEzhA0) said 
Because you use star svg that already has rounded angles. 
Just be attentive. Not all styles that you get at figma you need to copy into your css

src/styles/card.css
Viktor-Kost: It's not required yet, it's more as recommendation here. You will definitely learn about transition later. 
If you are still interested in - read [`this`](https://developer.mozilla.org/en-US/docs/Web/CSS/transition)

Viktor-Kost (approved):
Approved
Just read answers for your questions, and, you can apply recommendations.

src/styles/card.css
Viktor-Kost: It means that you will get the same width using right properties inside the card. Because it's not good when you hardcode height/width of an element wrapper like .card

src/styles/stars.css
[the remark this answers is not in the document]
Vasyl-Pavlenko: understood.
 thanks

PR #2234, opened 2022-12-02
Develop

GUSILLUS:
- [DEMO LINK](https://GUSILLUS.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://GUSILLUS.github.io/layout_product-cards/report/html_report/)

And can you explain me why i cant use `align-contents` to replace the margins (in code) with it?


src/style.css
```diff
@@ -0,0 +1,108 @@
...
+.card__image {
+  display: flex;
+  width: 160px;
+  height: 134px;
+  margin: auto;
+  margin-bottom: 40px;
+}
+
+.card__name {
+  display: flex;
```
OlEzhA0, line 28: why do we need `display: flex` here?

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  line-height: 18px;
+  color: var(--primary-grey);
+  margin-bottom: 4px;
+}
+
+.card__info {
+  display: inline-flex;
+  align-self: flex-start;
+  flex-wrap: wrap;
+  font-weight: 400;
```
OlEzhA0, line 40: why do we need this styles?

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  font-size: 10px;
+  line-height: 14px;
+  color: var(--secondary-grey);
+  margin-bottom: 16px;
+}
+
+.card__review {
+  display: flex;
+  width: 100%;
```
OlEzhA0, line 49: `display: flex` stretching for full width

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  align-items: center;
+  justify-content: space-between;
+  margin-bottom: 24px;
+}
+
+.card__review_number {
+  font-weight: 400;
```
OlEzhA0, line 56: it's redundant

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  font-size: 10px;
+  line-height: 14px;
+  display: flex;
```
OlEzhA0, line 59: we don't need this

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  color: var(--primary-grey);
+}
+
+.card__price {
+  display: flex;
+  width: 100%;
```
OlEzhA0, line 65: the same as I mentioned above

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  align-items: center;
+  justify-content: space-between;
+  margin-bottom: 16px;
+}
+
+.card__price_text {
+  font-weight: 400;
```
OlEzhA0, line 72: it's default as well

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  font-size: 16px;
+  line-height: 18px;
+  color: var(--primary-grey);
+}
+
+.card__link {
+  display: flex;
+  background: var(--primary-blue);
+  border-radius: 5px;
+  width: 166px;
```
OlEzhA0, line 89: we don’t need to define width here

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  height: 40px;
+  justify-content: center;
+  align-items: center;
+  text-align: center;
```
OlEzhA0, line 93: it's redundant

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  cursor: pointer;
```
OlEzhA0, line 94: it's default for `<a>`

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  font-size: 14px;
+  line-height: 16px;
+  text-transform: uppercase;
+  text-decoration: none;
+  color: #fff;
+}
+
+.card__link:hover {
+  background: #fff;
+  border: 1px solid var(--primary-blue);
```
OlEzhA0, line 105: you can define border in `.card__link` directly and don’t need to change it on hover

src/style.css
```diff
@@ -0,0 +1,108 @@
...
+  border-radius: 5px;
```
OlEzhA0, line 106: the same, why do we need change border-radius here?

src/vars.css
```diff
@@ -0,0 +1,5 @@
+:root {
+  --primary-blue: #00acdc;
```
witflash, line 2: It's better to add prefix for different variable types:
```
--color-primary-blue
--color-secondary-gray
...
--font-size-500
...
// etc.
```

OlEzhA0 (changes requested):
Hi! Good job! Pay attention to comments

PR #2269, opened 2022-12-07
Created a page with product card

svitjojo:


src/index.html
```diff
@@ -1,13 +1,69 @@
...
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+  </head>
+  <body>
+    <div class="card" data-qa="card">
+      <div class="card__image-box">
+        <img
+          class="card__image"
+          src="images/imac.jpeg"
+          alt="APPLE A1419 iMac 27 Retina 5K Monoblock"
+          >
```
Anna-Tkachenko, line 24:
```suggestion
        >
```

src/index.html
```diff
@@ -1,13 +1,69 @@
...
+      </div>
+
+      <div class="card__text">
+        <p class="card__title">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </p>
```
Anna-Tkachenko, line 30: will be better to use h2 here instead of p, it's title

src/index.html
```diff
@@ -1,13 +1,69 @@
...
+
+      <div class="card__reviews">
+        <div class="stars stars__stars--4">
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+        </div>
+        <div class="card__review-count">
```
Anna-Tkachenko, line 44: use semantic tags

src/index.html
```diff
@@ -1,13 +1,69 @@
...
+      <div class="card__price">
+        <p class="card__price-text">
+          Price:
+        </p>
+        <p class="card__price-amount">
+          $2,199
+        </p>
+      </div>
+
+      <div class="card__button">
```
Anna-Tkachenko, line 58: do you need the wrapper for the link? can we add all needed styles to the a element?

Anna-Tkachenko (changes requested):

src/index.html
```diff
@@ -1,13 +1,67 @@
...
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="anonymous"
+    >
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+  </head>
+  <body>
+    <div class="card" data-qa="card">
+      <div class="card__image-box">
```
Anna-Tkachenko, line 19: do you need additional wrapper for the image? can you set all needed properties to the image?

src/index.html
```diff
@@ -1,13 +1,67 @@
...
+        <img
+          class="card__image"
+          src="images/imac.jpeg"
+          alt="APPLE A1419 iMac 27 Retina 5K Monoblock"
+        >
+      </div>
+
+      <div class="card__text">
```
Anna-Tkachenko, line 27: same here, do you need additional wrapper `card__text`? try to set needed styles for the title and paragraph and delete redundant wrapper

src/index.html
```diff
@@ -1,13 +1,67 @@
...
+
+      <section class="card__reviews">
+        <div class="stars stars__stars--4">
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+        </div>
+        <div class="card__review-count">
```
Anna-Tkachenko, line 44: you didn't fix this comment, use semantic tag instead of div

src/style/badreset.css
```diff
@@ -0,0 +1,6 @@
+* {
+  font-family: "Roboto", "Segoe UI", sans-serif;
+  margin: 0;
+  padding: 0;
```
Anna-Tkachenko, line 4: don't reset margins and paddings using * selector
reset only needed properties for needed elements

src/style/badreset.css
```diff
@@ -0,0 +1,6 @@
+* {
+  font-family: "Roboto", "Segoe UI", sans-serif;
```
Anna-Tkachenko, line 2: move font styles to html tag

Anna-Tkachenko (changes requested):

src/index.html
svitjojo: @Anna-Tkachenko I used a wrapper to prevent cases if we have some problem with img, then a wrapper will still keep place and doesn't break the card.

Denchepornyuk: you don't need wrapper for this case, it's better to handle such cases in css, just add need styles for image and this will work for every image in cards

src/index.html
```diff
@@ -1,13 +1,66 @@
...
+
+      <h2 class="card__title">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </h2>
+      
+      <p class="card__product-code">
+          Product code: 195434
+      </p>
+
+      <section class="card__reviews">
```
Denchepornyuk, line 35: it isn't a section, section should have title and block with text, for stars it's better to use `div`

src/index.html
```diff
@@ -1,13 +1,66 @@
...
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="anonymous"
+    >
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+  </head>
+  <body>
+    <div class="card" data-qa="card">
+      <div class="card__image-box">
```
Denchepornyuk, line 19: it's better to use article tag for a card

src/style/badreset.css
```diff
@@ -0,0 +1,11 @@
+html {
+  font-family: "Roboto", "Segoe UI", sans-serif;
+}
+
+body,
+h2,
+p,
+div {
+  margin: 0;
+  box-sizing: border-box;
```
Denchepornyuk, line 10: box-sizing shouldn't be with reset styles, add  box-sizing to elements where you need it

Denchepornyuk (changes requested):

Denchepornyuk (approved):
good job 👍

PR #2284, opened 2022-12-09
Performed the task

max-rozzhalovets:
- [DEMO LINK](https://maasiimka.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://maasiimka.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -14,7 +14,7 @@
     <link rel="stylesheet" href="./stars.css">
   </head>
   <body>
-    <div class="product-card" data-qa="card">
+    <article class="product-card" data-qa="card">
       <img
         class="product-card__image"
         src="./images/imac.jpeg"
@@ -22,13 +22,15 @@
       >
       <h2 class="product-card__name">APPLE A1419 iMac 27&quot; Retina 5K Monoblock (MNED2UA/A)</h2>
       <p class="product-card__code">Product code: 195434</p>
-      <div class="stars stars--4">
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <p class="stars__reviews">Reviews: 5</p>
+      <div class="product-card__rating-container">
+        <div class="stars stars--4">
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+        </div>
+        <p class="product-card__reviews">Reviews: 5</p>
       </div>
       <div class="product-card__price-container">
         <p class="product-card__text">Price</p>
@@ -41,6 +43,6 @@
       >
         buy
       </a>
-    </div>
+    </article>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -22,8 +22,7 @@ p {
 
 .product-card__image {
   display: block;
-  margin-left: auto;
-  margin-right: auto;
+  margin: 0 auto;
   width: 160px;
   height: 134px;
 }
@@ -44,6 +43,21 @@ p {
   color: var(--secondary-color);
 
   padding-top: 4px;
+}
+
+.product-card__rating-container {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+
+  padding-top: 16px;
+}
+
+.product-card__reviews {
+  font-weight: 400;
+  font-size: 10px;
+  line-height: 14px;
+  color: var(--main-accent-color);
 }
 
 .product-card__price-container {
--- a/src/stars.css
+++ b/src/stars.css
@@ -1,6 +1,5 @@
 .stars {
   display: flex;
-  padding-top: 16px;
   align-items: end;
 }
 
@@ -11,15 +10,6 @@
   background-image: url(./images/star.svg);
   background-position: center;
   background-repeat: no-repeat;
-}
-
-.stars__reviews {
-  font-weight: 400;
-  font-size: 10px;
-  line-height: 14px;
-  color: var(--main-accent-color);
-
-  margin-left: 17px;
 }
 
 .stars--5 > .stars__star,


src/index.html
```diff
@@ -1,13 +1,43 @@
...
-  <h1>Product cards</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
+    <title>Product cards</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com"><link rel="preconnect" href="https://fonts.gstatic.com"><link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
```
GarryRocksman, line 10: this line is too long, you can separate it. One link one line.

src/index.html
```diff
@@ -1,13 +1,43 @@
...
+    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="./stars.css">
+  </head>
+  <body>
+    <div class="product-card" data-qa="card">
+      <img
+        class="product-card__image"
+        src="./images/imac.jpeg"
+        alt="APPLE A1419 iMac 27 Retina 5K Monoblock (MNED2UA/A)">
```
GarryRocksman, line 19: closing > at new line

maasiimka: Thanks, i missed that

src/style.css
```diff
@@ -0,0 +1,97 @@
...
+  box-sizing: border-box;
+  padding: 32px 16px 16px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+
+.product-card__image {
+  display: block;
+  margin-left: auto;
+  margin-right: auto;
```
GarryRocksman, line 26: you won't need this positioning if you will use display flex for the parent element. Here you just need align-self: center after all.

src/index.html
```diff
@@ -1,13 +1,43 @@
...
+      <h2 class="product-card__name">APPLE A1419 iMac 27&quot; Retina 5K Monoblock (MNED2UA/A)</h2>
+      <p class="product-card__code">Product code: 195434</p>
+      <div class="stars stars--4">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <p class="stars__reviews">Reviews: 5</p>
+      </div>
```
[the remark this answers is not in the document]
maasiimka: I don't think so. If I add div without class, I can't style it. Also, if I add div with class determined as BEM element, it violates rules of BEM (BEM block can't be inserted inside BEM element). If I add div with class determined as BEM block, it violates rules of task (there should be only stars and card blocks). But I'm not sure about that for 100%.  Anyway, thank you for the review.

src/style.css
```diff
@@ -0,0 +1,97 @@
...
+  border-radius: 5px;
+
+  transition: color 300ms, background-color 300ms, border-color 300ms;
+}
+
+.product-card:hover .product-card__button {
+  color: var(--blue-accent);
+  background-color: #fff;
+  border-color: var(--blue-accent);
+}
```
GarryRocksman, line 97: here I don't understand why you put hover on all product cards. I think we need to do it just for the button `product-card__button: hover ` will be enough

GarryRocksman (changes requested):
You did great work! But need some improvements.  Just open a checklist and revise one more time if all your code is written according to the rules

src/index.html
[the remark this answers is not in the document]
GarryRocksman: but you modified BEM  block and its also wrong )))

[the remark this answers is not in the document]
maasiimka: I dont see the rule that forbids it. I added p, but didnt change selectors and style rules for stars. As I understand it is the same as add p as a neighbor element

src/style.css
```diff
@@ -0,0 +1,97 @@
...
+
+}
+
+.product-card {
+  width: 200px;
+  box-sizing: border-box;
+  padding: 32px 16px 16px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
```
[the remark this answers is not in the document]
maasiimka: Is flex needed there? Three more lines for display:flex, flex-direction: column and align-items:center, but the same result

src/index.html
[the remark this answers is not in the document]
GarryRocksman: ok, let's see what the mentor will say, because i inerested to. as i wrote maybe I'm wrong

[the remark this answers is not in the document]
maasiimka: I do not claim that I'm right. One way or another we will see ;)

src/style.css
maasiimka: I could have done it, but in the comment higher I explained why not. But your idea seems good, thanks for it

src/style.css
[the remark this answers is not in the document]
GarryRocksman: that's my opinion ))) yes you adding a couple of lines here but else elements you can position with one line )

src/style.css
maasiimka: Agree with you. I confused it with the case when there is no button and the whole block is link.

src/style.css
```diff
@@ -0,0 +1,97 @@
...
+  height: 134px;
+}
+
+.product-card__name {
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 18px;
+  color: var(--main-accent-color);
+
+  padding-top: 40px;
```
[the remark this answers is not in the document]
maasiimka: To tell the truth, not sure that it plays any role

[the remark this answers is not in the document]
GarryRocksman: maybe ) as I wrote  - I can be wrong  )))

src/index.html
```diff
@@ -1,13 +1,46 @@
...
+    >
+    <title>Product cards</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="./stars.css">
+  </head>
+  <body>
+    <div class="product-card" data-qa="card">
```
Denchepornyuk, line 17: it's better to use article tag for a card

src/index.html
```diff
@@ -1,13 +1,46 @@
...
+      <h2 class="product-card__name">APPLE A1419 iMac 27&quot; Retina 5K Monoblock (MNED2UA/A)</h2>
+      <p class="product-card__code">Product code: 195434</p>
+      <div class="stars stars--4">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <p class="stars__reviews">Reviews: 5</p>
+      </div>
```
Denchepornyuk, line 32: in this case, you can add a modifier for the `stars` block, it will mean that the block has a state with 4 active stars

src/index.html
```diff
@@ -1,13 +1,46 @@
...
+      >
+      <h2 class="product-card__name">APPLE A1419 iMac 27&quot; Retina 5K Monoblock (MNED2UA/A)</h2>
+      <p class="product-card__code">Product code: 195434</p>
+      <div class="stars stars--4">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <p class="stars__reviews">Reviews: 5</p>
```
Denchepornyuk, line 31: reviews shouldn't be in the stars block, move it to the product-card and connect reviews to product-card - `product-card__reviews`

src/style.css
Denchepornyuk, line 26: you can use one css rule instead of two
margin: 0 auto;

Denchepornyuk (changes requested):

src/index.html
maasiimka: Maybe I didn't understand you, but isn't modifier already added for stars block?

src/index.html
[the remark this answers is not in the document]
maasiimka: You were right, I shouldn't have modified stars block. Thanks for this discussion ;)

Denchepornyuk (approved):

PR #2346, opened 2022-12-18
solution

vladis9:
DEMO LINK (https://vladis9.github.io/layout_product-cards/)
TEST REPORT LINK (https://vladis9.github.io/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,57 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <title>Product cards</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Product cards</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <title>Product cards</title>
+    <link rel="stylesheet" href="style.css">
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,129 @@
...
+  height: 16px;
+  margin-right: 4px;
+  background-repeat: no-repeat;
+  background-position: center;
+  background-image: url(images/star.svg);
+}
+
+.stars__star:last-child {
+  margin: 0;
+}
```
tetiana-ivanchenko, line 71:
```suggestion
  background-repeat: no-repeat;
  background-position: center;
  background-image: url(images/star.svg);
}

.stars__star:not(:last-child) {
  margin-rignt: 4px;
}
```

tetiana-ivanchenko (changes requested):

src/style.css
```diff
@@ -0,0 +1,128 @@
...
+  font-style: normal;
+  font-size: 12px;
+  line-height: 18px;
+  color: var(--color-dark);
+  margin-bottom: 4px;
+}
+
+.card__subtitle {
+  font-style: normal;
+  font-weight: 400;
```
TarasHoliuk, line 41: default values, so no need to set it (FYI)

src/index.html
```diff
@@ -1,13 +1,61 @@
...
+          $2,199
+        </div>
+      </div>
+
+      <a
+        href="#"
+        class="card__buy"
+        data-qa="hover"
+        >
+          BUY
```
TarasHoliuk, line 57: it's better to set `text-transform: uppercase` in css instead of write `BUY` (FYI)

TarasHoliuk (approved):
GJ! I've left 2 comments just FYI - no need to push changes (PR is already approved)

src/style.css
vladis9: exactly :)

src/index.html
vladis9: yes, thanks )

PR #2442, opened 2023-01-21
Solution

Ihor-Zakharov:


src/style.css
```diff
@@ -0,0 +1,121 @@
...
+* {
+  box-sizing: border-box;
+  margin: 0;
+}
+
+.card {
+  display: flex;
+  flex-direction: column;
+  width: 200px;
+  /* background-color: yellowgreen; */
```
SanyaBratashchuk, line 12: Do not leave comments

src/style.css
```diff
@@ -0,0 +1,121 @@
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap");
+
+* {
+  box-sizing: border-box;
+  margin: 0;
```
SanyaBratashchuk, line 5: Do not apply any rules to the `*` selector except `box-sizing`

src/style.css
```diff
@@ -0,0 +1,121 @@
...
+}
+
+.card {
+  display: flex;
+  flex-direction: column;
+  width: 200px;
+  /* background-color: yellowgreen; */
+  padding: 32px 16px 28px;
+  font-family: "Roboto", sans-serif;
```
SanyaBratashchuk, line 14: Move font-family to the html tag selector

src/style.css
```diff
@@ -0,0 +1,121 @@
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap");
+
+* {
+  box-sizing: border-box;
+  margin: 0;
+}
+
+.card {
```
SanyaBratashchuk, line 8: Move each BEM block to the separate file

src/style.css
```diff
@@ -0,0 +1,121 @@
...
+  background-image: url(images/star.svg);
+}
+
+.stars > .star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars--lighted_4 > div:nth-child(-n + 4) {
+  background-image: url(images/star-active.svg);
+}
```
SanyaBratashchuk, line 79: Reuse all styles from task with stars

SanyaBratashchuk (changes requested):

ViktorHavryliuk (changes requested):
Please check if you have pushed correct changes, as now style.css is deleted and there is no demo link

SanyaBratashchuk (changes requested):

src/index.html
```diff
@@ -1,13 +1,58 @@
...
+    <div class="card" data-qa="card">
+      <img 
+        src="images/imac.jpeg"
+        alt="imac"
+        class="card__img"
+      >
+
+      <h1 class="card__name">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </h1>
```
SanyaBratashchuk, line 26: Use h3 better. Because h1 can be only one on the page. But you can have more than one card.

src/styles/blocks/stars.css
```diff
@@ -0,0 +1,24 @@
+.stars {
+  display: flex;
+  flex-direction: row;
```
SanyaBratashchuk, line 3:
```suggestion
```
Default  value

SanyaBratashchuk (changes requested):

tetiana-ivanchenko (approved):

PR #2444, opened 2023-01-22
Develop

LeoSpartaDeV:
https://LeoSpartaDeV.github.io/layout_product-cards
https://LeoSpartaDev.github.io/layout_product-cards/report/html_report


src/index.html
```diff
@@ -1,13 +1,71 @@
...
+        <div class="stars stars--4 card__rating-stars">
+          <div class="stars__star"></div>
+
+          <div class="stars__star"></div>
+
+          <div class="stars__star"></div>
+
+          <div class="stars__star"></div>
+
+          <div class="stars__star"></div>
```
Volodymyr-Melnyk-Fron10der, line 44: Split your blocks by new line it is a good practice and I like it in your html), BUT in this case I would recommend don't do this, because those tags are one-line and easy to read even when they are placed line by line

src/index.html
```diff
@@ -1,13 +1,71 @@
...
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <title>Product cards</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Product cards</h1>
-</body>
+  <head class="page">
```
Volodymyr-Melnyk-Fron10der, line 3: head doesn't affect on your layout, so I think there is no reason to style it

```suggestion
  <head>
```

Volodymyr-Melnyk-Fron10der (changes requested):
Excellent work! I like it but there are a few improvments:

1. Split your styles into different files (example: card.css, start.css, main.css, ...)

src/css/fonts.css
```diff
@@ -0,0 +1,20 @@
+@font-face {
+  font-family: Roboto-regular;
+  src: url(../fonts/Roboto-Regular.ttf);
+  font-weight: 400;
+  font-style: normal;
```
Volodymyr-Melnyk-Fron10der, line 5: you don't have to set default values

Volodymyr-Melnyk-Fron10der (approved):
MichaelScottGIF

PR #2490, opened 2023-01-29
Solution

Taras-1:
- [DEMO LINK](https://Taras-1.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://Taras-1.github.io/layout_product-cards/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,146 @@
...
+  background-color: rgb(159, 151, 141);
+}
+
+.card {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  width: 200px;
+  box-sizing: border-box;
```
TarasHoliuk, line 12: it should be in `*` selector

src/style.css
```diff
@@ -0,0 +1,146 @@
...
+  margin-top: 32px;
+}
+
+.card__description {
+  margin-top: 40px;
+  width: 166px;
+  height: 36px;
+  font-size: 12px;
+  line-height: 18px;
+  font-style: normal;
```
TarasHoliuk, line 30: default value - no need to set it

src/style.css
```diff
@@ -0,0 +1,146 @@
...
+  font-weight: 500;
+  color: #060b35;
+}
+
+.card__product-code {
+  margin-top: 4px;
+  align-self: flex-start;
+  font-size: 10px;
+  line-height: 14px;
+  font-weight: 400;
```
TarasHoliuk, line 40: default value

src/style.css
```diff
@@ -0,0 +1,146 @@
...
+  font-style: normal;
+  color: #616070;
+}
+
+.stars-container {
+  display: flex;
+  flex-direction: row;
```
TarasHoliuk, line 47: default value

TarasHoliuk (changes requested):
GJ :+1: 
Pls, delete all default css values

src/style.css
```diff
@@ -0,0 +1,138 @@
+* {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  box-sizing: border-box;
```
vlad-demchuk, line 5: it meant only `box-sizing` should be in `*`. return other properties.

src/style.css
vlad-demchuk: fix it

vlad-demchuk (changes requested):
fix comments

vlad-demchuk (changes requested):
fix the previous comment
`font-weight: 400` is a default value
image

vlad-demchuk (approved):

PR #2497, opened 2023-01-29
Develop

yuliia-vereshko:
- [DEMO LINK](https://yuliia-vereshko.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://yuliia-vereshko.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/styles/card.css
+++ b/src/styles/card.css
@@ -1,0 +1,81 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  width: 200px;
+  padding: 32px 16px 16px;
+
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+
+.card__image {
+  margin: 0 auto 40px;
+  width: 160px;
+  height: 134px;
+}
+
+.card__title {
+  font-size: 12px;
+  font-weight: 500;
+  line-height: 18px;
+  margin-bottom: 4px;
+  color: #060b35;
+}
+
+.card__code {
+  color: #616070;
+  margin-bottom: 16px;
+  line-height: 14px;
+}
+
+.card__rating {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin-bottom: 24px;
+}
+
+.card__rating-review {
+  color: #060b35;
+}
+
+.card__price {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin-bottom: 16px;
+}
+
+.card__price-text {
+  color: #616070;
+  font-size: 12px;
+  line-height: 18px;
+}
+
+.card__price-value {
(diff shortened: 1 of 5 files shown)


src/styles/card.css
```diff
@@ -0,0 +1,81 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  width: 200px;
+  padding: 32px 16px 16px;
+
+  border: 1px solid #f3f3f3;
```
stanislavkulakovskyi, line 7: not sure if it's necessary, but you could assign colours to a variables and using them that way. but again, there's no issue, just a suggestion

src/styles/card.css
```diff
@@ -0,0 +1,81 @@
...
+  width: 160px;
+  height: 134px;
+}
+
+.card__title {
+  font-size: 12px;
+  font-weight: 500;
+  line-height: 18px;
+  margin-bottom: 4px;
+  color: #060b35;
```
stanislavkulakovskyi, line 22: not sure if it's necessary, but you could assign colours to a variables and using them that way. but again, there's no issue, just a suggestion

src/styles/card.css
```diff
@@ -0,0 +1,81 @@
...
+  width: 200px;
+  padding: 32px 16px 16px;
+
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+
+.card__image {
+  margin: 0 auto 40px;
+  width: 160px;
```
stanislavkulakovskyi, line 13: not crucial, but probably it's better to use margin-bottom and also you can only specify height for an image

stanislavkulakovskyi (approved):
great job overall! just a few suggestions :)

src/styles/card.css
yuliia-vereshko: Thanks, I fixed width and set align-items:center, but when I use margin-bottom only, my image will be large than in mockup

src/styles/stars.css
```diff
@@ -0,0 +1,27 @@
+.stars {
+  display: flex;
+  margin: 0;
+  padding: 0;
```
illiarerun, line 4: divs don't have margins and paddings by default
```suggestion
```

src/styles/stars.css
```diff
@@ -0,0 +1,27 @@
...
+  justify-content: flex-start;
```
illiarerun, line 5: default value
```suggestion
```

src/styles/card.css
illiarerun: agree, better use variables for commonly used values, such as colors, please do it in this task

illiarerun (changes requested):
GJ, but few fixes left:

PR #2498, opened 2023-01-29
created card

valeriiayarzhemska:
- [DEMO LINK](https://valeriiayarzhemska.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://valeriiayarzhemska.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/styles/blocks/card.css
+++ b/src/styles/blocks/card.css
@@ -1,0 +1,119 @@
+.card {
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+
+  max-width: 200px;
+  padding: 0 16px 16px;
+
+  background-color: #fff;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+
+.card__image-container {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  margin-top: 32px;
+}
+
+.card__image {
+  width: 160px;
+  height: 134px;
+}
+
+.title {
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 18px;
+
+  color: var(--main-text-color);
+}
+
+.card__title {
+  margin: 40px 0 0;
+}
+
+.card__product-code {
+  margin: 4px 0 0;
+}
+
+.product-code {
+  font-size: 10px;
+  line-height: 14px;
+  font-weight: 400;
+
+  color: var(--secondary-text-color);
+}
+
+.card__paragraph {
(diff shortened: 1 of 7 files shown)


src/styles/blocks/card.css
```diff
@@ -0,0 +1,119 @@
+.card {
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+
+  max-width: 200px;
+  padding: 0 16px 16px;
```
anna-lymorenko, line 7: maybe we can set padding in this way - padding: 16px;

src/styles/blocks/card.css
```diff
@@ -0,0 +1,119 @@
...
+
+  background-color: #fff;
```
anna-lymorenko, line 9: we can use var for background-color

anna-lymorenko (commented):
i like your code, didn't find any serious issues ;)

src/styles/blocks/card.css
valeriiayarzhemska: > we can use var for background-color

we use it one time, so think dont need to make variable, probably

src/styles/blocks/card.css
valeriiayarzhemska: > maybe we can set padding in this way - padding: 16px;

need to do it like this to have correct paddings

src/styles/blocks/card.css
```diff
@@ -0,0 +1,119 @@
...
+  text-decoration: none;
+  color: #fff;
+
+  background-color: var(--blue-accent-color);
+  border: 1px solid var(--blue-accent-color);
+  border-radius: 5px;
+}
+
+.card__buy-button:hover {
+  transition: all 0.5s ease-out;
```
illiarerun, line 116: don't add transition only to hover rule, it will not work proper way, move it to `.card__buy-button`

illiarerun (changes requested):

src/styles/blocks/stars.css
```diff
@@ -0,0 +1,21 @@
...
+  background-position: 50%;
+  height: 16px;
+  width: 16px;
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars--4 :not(:last-child) {
```
illiarerun, line 19: i don't see modifiers for 1, 2, 3 and five active stars

illiarerun (approved):

PR #2553, opened 2023-02-09
added product-card

tatiana-sadaeva:
[DEMO LINK](https://tatiana-sadaeva.github.io/layout_product-cards/)
[TEST REPORT LINK](https://tatiana-sadaeva.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,133 @@
+:root {
+  --prim-text-color: #060b35;
+  --second-text-color: #616070;
+  --link-button-color: #00acdc;
+  --white-color: #fff;
+}
+
+* {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,133 @@
...
+}
+
+body {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+}
+
+.card {
```
SanyaBratashchuk, line 29: move each BEM block to the separate file

SanyaBratashchuk (changes requested):

PR #2570, opened 2023-02-13
Product-card-1

Soi4An:
DEMO LINK - https://Soi4An.github.io/layout_product-cards/
TEST REPORT LINK - https://Soi4An.github.io/layout_product-cards/report/html_report/


src/styles/block-card.css
```diff
@@ -0,0 +1,97 @@
...
+  margin: 0 0 40px;
+}
+
+.card__img {
+  width: 100%;
+  height: 134px;
+}
+
+.card__name {
+  font-family: Roboto, sans-serif;
```
YegorKorenkov, line 27: Please move this rule to body selector. You can declare this rule once and then not write it every time for all elements. Fix for all selectors!
And also it's not only for font-family, you can also declare default value of font-weight, line-height etc. in body for prevent reuse of this rules

src/styles/block-card.css
```diff
@@ -0,0 +1,97 @@
...
+.card__reviews,
+.card__text-price {
+  font-family: Roboto, sans-serif;
+  font-weight: 400;
+}
+
+.card__code {
+  font-size: 10px;
+  line-height: 14px;
+  columns: #616070;
```
YegorKorenkov, line 44: It's rule incorrect. I don't know why you add this. Please delete. Maybe you want to add color rule?

src/styles/block-card.css
```diff
@@ -0,0 +1,97 @@
...
+.card {
+  display: flex;
+  width: 200px;
+  padding: 32px 16px 16px;
+  box-sizing: border-box;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  flex-direction: column;
+  color: #060b35;
+  --button-color: #00acdc;
```
YegorKorenkov, line 11: More corretcly declare variables in _:root_ selector

src/styles/block-card.css
```diff
@@ -0,0 +1,97 @@
...
+.card__price {
+  display: flex;
+  justify-content: space-between;
+  margin: 0 0 16px;
+}
+
+.card__text-price {
+  font-size: 12px;
+  line-height: 18px;
+  color: #616070;
```
YegorKorenkov, line 68: You can also declare var with this color, because you use thi color in different places

src/styles/block-card.css
```diff
@@ -0,0 +1,97 @@
...
+  font-weight: 700;
+  font-size: 14px;
+  line-height: var(--button-height);
+  text-align: center;
+  border: 1px solid var(--button-color);
+}
+
+.card:hover .card__button {
+  color: var(--button-color);
+  background-color: #fff;
```
YegorKorenkov, line 95: You can move this color to var also. If you added other colors to the variables, then add this one too, because if you want to change the color, you will have many buttons in different places, you won’t need to search through all the styles to change them, but it will be enough to change only the one variable.

src/styles/block-card.css
Soi4An: it's color) it's work of Tab..

src/styles/block-card.css
Soi4An: in my opinion, these colors may to do different

src/styles/block-card.css
Soi4An: @YegorKorenkov But if i will want to transfer this block, these variables must to be inside the block .card

src/styles/block-card.css
Soi4An: again, in my opinion, these colors may to do different. But okey

src/styles/block-card.css
```diff
@@ -0,0 +1,90 @@
...
+.card__reviews,
+.card__text-price {
+  font-weight: 400;
+}
+
+.card__code {
+  font-size: 10px;
+  line-height: 14px;
+  color: var(--color-text-code-price);
+  margin: 0 0 16px;
```
vlad-demchuk, line 40: try to set only `margin-bottom` inside elements. if some elements have default margins, you can just reset it for tag, not need to do it for each class
image

src/styles/block-card.css
```diff
@@ -0,0 +1,90 @@
...
+  font-weight: 700;
+  font-size: 16px;
+  line-height: 18px;
+}
+
+.card__button {
+  height: var(--button-height);
+  color: var(--button-color-text-bgc-hover);
+  background-color: var(--button-color);
+  box-sizing: border-box;
```
vlad-demchuk, line 76: you can just move `box-sizing: border-box` to `*` selector and no need to repeat it then

src/styles/block-card.css
```diff
@@ -0,0 +1,90 @@
...
+  border-radius: 5px;
+  text-decoration: none;
+  font-weight: 700;
+  font-size: 14px;
+  line-height: var(--button-height);
+  text-align: center;
+  border: 1px solid var(--button-color);
+}
+
+.card:hover .card__button {
```
vlad-demchuk, line 86: use hover only for button

src/styles/reset.css
```diff
@@ -0,0 +1,15 @@
+/* #region reset */
+body {
+  margin: 0;
+  font-family: Roboto, sans-serif;
+  color: #060b35;
+  --button-color: #00acdc;
+  --button-height: 40px;
+  --button-color-text-bgc-hover: #fff;
+  --color-text-code-price: #616070;
```
vlad-demchuk, line 9: move variable to `:root` pseudo selector

vlad-demchuk (changes requested):
If you have any questions, feel free to ask in `fe_chat`

slavanosatskyi (approved):
Looks good 👍

PR #2617, opened 2023-02-24
Develop

vinnykalisa:
- [DEMO LINK](https://vinnykalisa.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://vinnykalisa.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/blocks/card.css
+++ b/src/blocks/card.css
@@ -69,8 +69,7 @@
 
 .card__buy-link {
   display: flex;
-  margin-top: 16px;
-  margin-bottom: 16px;
+  margin: 16px 0;
   justify-content: center;
   align-items: center;
   cursor: pointer;
@@ -83,12 +82,12 @@
   line-height: 16px;
   color: #fff;
   background: #00acdc;
-  border: 1px solid #f3f3f3;
+  border: 1px solid transparent;
   border-radius: 5px;
 }
 
 .card__buy-link:hover {
   background-color: #fff;
-  border: 1px solid transparent;
+  border: 1px solid #00acdc;
   color: #00acdc;
 }


src/style.css
```diff
@@ -0,0 +1,22 @@
+@import "./blocks/card.css";
+@import "./blocks/stars.css";
+
+* {
+  box-sizing: border-box;
+}
+
+body,
+h2,
+h3,
```
prokivskaaa: maybe here it is better to use classes? According to the checklist p1 => `Get used to style all elements using classes. And don't increase selectors specificity unless completely necessary.`

vinnykalisa: but in lectures Misha did this in reset file:
body, h1, h2, h3, ul, p {
  margin: 0;
}

prokivskaaa (changes requested):
Great. I have some comments. Please, have a look, maybe some of them are useful

src/style.css
prokivskaaa: @vinnykalisa So it is better to ask the mentor as a checklist seems a little controversial with Misha's lectures. Maybe it is an exeption

src/blocks/card.css
```diff
@@ -0,0 +1,94 @@
...
+.card__price-number {
+  font-weight: 700;
+  font-size: 16px;
+  line-height: 18px;
+}
+
+.card__buy-link {
+  display: flex;
+  margin-top: 16px;
+  margin-bottom: 16px;
```
prokivskaaa, line 73:
```suggestion
  margin: 16px 0;
```

src/blocks/card.css
```diff
@@ -0,0 +1,94 @@
...
+  padding: 12px;
+  text-decoration: none;
+  text-transform: uppercase;
+  text-align: center;
+  font-weight: 700;
+  font-size: 14px;
+  line-height: 16px;
+  color: #fff;
+  background: #00acdc;
+  border: 1px solid #f3f3f3;
```
prokivskaaa, line 86:
```suggestion
  border: 1px solid transparent;
```

src/style.css
vinnykalisa: If you'll need it later - they told be that for reset we should use body, p, h2...

src/blocks/card.css
vinnykalisa: Thank you for review!

liliia-livandovskaya (approved):
🔥 🔥 🔥 Excellent 🔥 🔥 🔥

PR #2707, opened 2023-03-07
Product cards

Dorosh90:
- [DEMO LINK](https://Dorosh90.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://Dorosh90.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -23,26 +23,26 @@
         src="./images/imac.jpeg"
         alt="APPLE A1419 iMac 27 Retina 5K Monoblock (MNED2UA/A)"
       >
-      <section>
-        <h1>APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</h1>
+      <div class="card__info-product">
+        <p class="product-name">APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</p>
         <p class="product-code">Product code: 195434</p>
-      </section>
+      </div>
 
-      <section class="section-stars">
-        <div class="section-stars__section section-stars__section--star-4">
-          <div class="section-stars__star"></div>
-          <div class="section-stars__star"></div>
-          <div class="section-stars__star"></div>
-          <div class="section-stars__star"></div>
-          <div class="section-stars__star"></div>
+      <div class="card__stars">
+        <div class="card__stars-section card__stars-section--star-4">
+          <div class="card__stars-star"></div>
+          <div class="card__stars-star"></div>
+          <div class="card__stars-star"></div>
+          <div class="card__stars-star"></div>
+          <div class="card__stars-star"></div>
         </div>
-        <p class="section-stars__reviews">Reviews: 5</p>
-      </section>
+        <p class="card__stars-reviews">Reviews: 5</p>
+      </div>
 
-      <section class="section-price">
-        <p class="section-price__info">Price:</p>
-        <p class="section-price__price">$2,199</p>
-      </section>
+      <div class="card__price-block">
+        <p class="card__info">Price:</p>
+        <p class="card__price">$2,199</p>
+      </div>
 
       <button class="card__button" data-qa="hover">BUY</button>
     </div>
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,50 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <title>Product cards</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Product cards</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+      >
+    <title>Product cards</title>
+    <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700& display=swap"
+      rel="stylesheet"
+    >
+  </head>
+
+  <body>
+    <div class="card card--size" data-qa="card">
+      <img
+        class="card__logo"
+        src="./images/imac.jpeg"
+        alt="APPLE A1419 iMac 27 Retina 5K Monoblock (MNED2UA/A)"
+      >
+      <section>
+        <h1>APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</h1>
+        <p class="product-code">Product code: 195434</p>
+      </section>
+
+      <section class="section-stars">
+        <div class="section-stars__section section-stars__section--star-4">
+          <div class="section-stars__star"></div>
+          <div class="section-stars__star"></div>
+          <div class="section-stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -1,13 +1,50 @@
...
+    <div class="card card--size" data-qa="card">
+      <img
+        class="card__logo"
+        src="./images/imac.jpeg"
+        alt="APPLE A1419 iMac 27 Retina 5K Monoblock (MNED2UA/A)"
+      >
+      <section>
+        <h1>APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</h1>
+        <p class="product-code">Product code: 195434</p>
+      </section>
```
anstsot, line 29: it's better to use `div` instead of `section`. check other places as well.
The HTML Section element represents a generic section of a document, i.e., a thematic grouping of content, typically with a heading.
```suggestion
      <div>
        <h1>APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</h1>
        <p class="product-code">Product code: 195434</p>
      </div>
```

src/index.html
```diff
@@ -1,13 +1,50 @@
...
+
+  <body>
+    <div class="card card--size" data-qa="card">
+      <img
+        class="card__logo"
+        src="./images/imac.jpeg"
+        alt="APPLE A1419 iMac 27 Retina 5K Monoblock (MNED2UA/A)"
+      >
+      <section>
+        <h1>APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</h1>
```
anstsot, line 27: it's better not to use h1 element in card. what if you have more cards? the H1 title should be the only one on whole page.

src/style.css
```diff
@@ -0,0 +1,142 @@
...
+}
+
+.card__logo {
+  display: block;
+  margin: 0 auto 40px;
+  width: 160px;
+  height: 134px;
+}
+
+h1 {
```
anstsot, line 32: it's not good add styles to tag. please add class and use it here as selector.

src/style.css
```diff
@@ -0,0 +1,142 @@
...
+  font-family: Roboto, sans-serif;
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 18px;
+  color: var(--mainColor);
+  margin: 0 0 4px;
+}
+
+p {
```
anstsot, line 42: same here (please add class and use it here as selector.)

src/index.html
```diff
@@ -1,13 +1,50 @@
...
+        <p class="product-code">Product code: 195434</p>
+      </section>
+
+      <section class="section-stars">
```
anstsot, line 31: it would be better to use BEM methodology and change the class name to:
```suggestion
      <section class="card__stars">
```

src/index.html
```diff
@@ -1,13 +1,50 @@
...
+          <div class="section-stars__star"></div>
+          <div class="section-stars__star"></div>
+        </div>
+        <p class="section-stars__reviews">Reviews: 5</p>
+      </section>
+
+      <section class="section-price">
+        <p class="section-price__info">Price:</p>
+        <p class="section-price__price">$2,199</p>
+      </section>
```
anstsot, line 45: same here with class 
```suggestion
      <section class="card__price-block">
        <p class="card__info">Price:</p>
        <p class="card__price">$2,199</p>
      </section>
```

anstsot (commented):
Looks good, but there are few fixes to improve your code.

anstsot (changes requested):
Changes requested

src/style.css
```diff
@@ -0,0 +1,157 @@
...
+
+/* p {
+  font-family: Roboto, sans-serif;
+  font-style: normal;
+  font-weight: 400;
+  font-size: 10px;
+  line-height: 14px;
+  color: var(--mainColor);
+  margin: 0;
+} */
```
OleziO, line 23: Remove all comments from the code.

src/style.css
```diff
@@ -0,0 +1,157 @@
...
+
+:root {
+  --mainColor: #060b35;
+  --blueAccent: #00acdc;
+}
+
+.card {
+  background: #fff;
+  border: 1px solid #f3f3f3;
```
OleziO, line 32: Move this color code to a variable.

src/style.css
```diff
@@ -0,0 +1,157 @@
...
+  border-radius: 5px;
+  width: 200px;
+  box-sizing: border-box;
```
OleziO, line 35: Move this to `*` selector.

src/style.css
```diff
@@ -0,0 +1,157 @@
...
+
+.card__logo {
+  display: block;
+  margin: 0 auto 40px;
+  width: 160px;
+  height: 134px;
+}
+
+.product-name {
+  font-family: Roboto, sans-serif;
```
OleziO, line 52: Move this to `body` selector

src/style.css
```diff
@@ -0,0 +1,157 @@
...
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 18px;
+  color: var(--mainColor);
+  margin: 0 0 4px;
+}
+
+.product-code {
+  font-family: Roboto, sans-serif;
```
OleziO, line 62: Don't repeat this in the code after adding to the body selector. Check other similar cases.

src/style.css
OleziO: Check other similar cases.

src/style.css
```diff
@@ -0,0 +1,157 @@
...
+  background-position: center;
+  height: 16px;
+  width: 16px;
+}
+
+.card__stars-star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.card__stars-section--star-0 :nth-child(-n+0),
```
OleziO, line 94: Unnecessary line, it does not add any star, just remove it.

src/style.css
```diff
@@ -0,0 +1,157 @@
...
+.card__stars-section--star-1 :nth-child(-n+1),
+.card__stars-section--star-2 :nth-child(-n+2),
+.card__stars-section--star-3 :nth-child(-n+3),
+.card__stars-section--star-4 :nth-child(-n+4) {
+  background-image: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  height: 16px;
+  width: 16px;
```
OleziO, line 103: You don't need to set this in this class cause you already do it in the empty star class.

src/style.css
```diff
@@ -0,0 +1,157 @@
...
+}
+
+.card__stars-reviews {
+  font-family: Roboto, sans-serif;
```
OleziO, line 107: Remove this line.

src/style.css
```diff
@@ -0,0 +1,157 @@
...
+.card__price {
+  font-family: Roboto, sans-serif;
+  font-style: normal;
+  font-weight: 700;
+  font-size: 16px;
+  line-height: 18px;
+}
+
+.card__button {
+  margin: auto;
```
OleziO, line 138: The better practice is to set the margin to 0 auto; when you want to centralize the element.

OleziO (changes requested):

src/style.css
Dorosh90: All this code? it will add borders to all elements

OleziO: `box-sizing: border-box;` needed to display elements on the page correctly. It is good practice to add this property to all elements (using `*` selector)

OleziO (changes requested):

src/style.css
```diff
@@ -0,0 +1,136 @@
...
+
+.card__button {
+  margin: 0 auto;
+  color: var(--whiteAccent);
+  width: 166px;
+  height: 40px;
+  background-color: var(--blueAccent);
+  border: none;
+  border-radius: 5px;
+  font-family: Roboto, sans-serif;
```
OleziO, line 124: Remove this line. You don't need to specify the font family here.

OleziO (approved):

PR #2718, opened 2023-03-09
Develop

Zibi95:
[DEMO LINK](https://Zibi95.github.io/layout_product-cards/)
[TEST REPORT LINK](https://Zibi95.github.io/layout_product-cards/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,5 @@
+@import "./styles/variables.css";
+
+@import "./styles/reset.css";
+@import "./styles/stars.css";
+@import "./styles/card.css";
```
DorotaLeniecDev, line 5: This is a small project so using @import can help to organise your CSS, but you need to remember that there are also downsides of this approach:

- Performance impact: Using too many imports can slow down page load times. Each import requires a separate HTTP request, which can add up if there are many imports.
- Order of imports: The order in which stylesheets are imported can affect the way styles are applied to the page. This can create unexpected results if not managed properly.

src/styles/card.css
```diff
@@ -0,0 +1,86 @@
+.card {
+  font-family: "Roboto", Arial, sans-serif;
+  width: 200px;
+  padding-top: 32px;
+  padding-left: var(--horizontal-spacing);
+  padding-right: var(--horizontal-spacing);
+  padding-bottom: var(--vertical-spacing);
```
DorotaLeniecDev, line 7: This can be simplified to oneliner:
```suggestion
  padding: 32px var(--horizontal-spacing) var(--horizontal-spacing);
```

src/styles/card.css
```diff
@@ -0,0 +1,86 @@
...
+  border-radius: 5px;
+  border: 1px solid #f3f3f3;
+  display: flex;
+  flex-direction: column;
+}
+
+.card__product-img {
+  width: 160px;
+  height: 134px;
```
DorotaLeniecDev, line 16: There is no need to set it one more time, as you set **width** and **height** on img tag

src/styles/variables.css
```diff
@@ -0,0 +1,18 @@
+:root {
+  --main-font-color: #060b35;
+  --secondary-font-color: #616070;
+
+  --button-color: #00acdc;
```
DorotaLeniecDev, line 5:
```suggestion
  --button-blue: #00acdc;
```
It is good when variable are as much descriptive as they can, for better readability

src/styles/stars.css
```diff
@@ -0,0 +1,24 @@
...
+  margin-right: 4px;
+  background-position: center;
+  background-repeat: no-repeat;
+  width: 16px;
+  height: 16px;
+}
+
+.stars--5 .stars__star {
+  background-image: url(../images/star-active.svg);
+}
```
DorotaLeniecDev, line 17: You can merge this also to below selector :)

src/styles/variables.css
```diff
@@ -0,0 +1,18 @@
...
+
+  --font-size-small: 10px;
+  --font-size-medium: 12px;
+
+  --font-height-medium: 14px;
+  --font-height-large: 18px;
```
DorotaLeniecDev, line 11: For consistancy I would name those variable like this:
```suggestion
  --font-size-small: 10px;
  --font-size-medium: 12px;

  --font-height-small: 14px;
  --font-height-medium: 18px;
```

src/styles/variables.css
```diff
@@ -0,0 +1,18 @@
...
+
+  --font-weight-bold: 700;
+  --font-weight-normal: 400;
+
+  --horizontal-spacing: 16px;
+  --vertical-spacing: 16px;
```
DorotaLeniecDev, line 17: You could merge those 2 as they have the same value
```suggestion
  --spacing: 16px;
```

DorotaLeniecDev (changes requested):
Really good job! I like file structure as it is easy to read. Check my comments as they can improve your code a bit.

src/index.html
```diff
@@ -1,13 +1,64 @@
...
+        width="160"
+        height="134"
+        class="card__image"
+      >
+
+      <h2 class="card__title">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </h2>
+
+      <p class="card__code">Product code: 195434</p>
```
Radoslaw-Czerniawski, line 35: I'm not sure if I would use paragraph for such small pieces of text. Maybe `span` would be better, or you could wrap the code here in span. It will be easier for you to change it dynamically with javascript in the future as the code won't be static but rather configurable by the site's owner, same for the number part of the score.

Radoslaw-Czerniawski (approved):
Left some feedback but generally a very good pr.

src/index.html
```diff
@@ -1,13 +1,64 @@
...
+
+      <div class="review card__review">
```
DorotaLeniecDev, line 37: From Checklist: _Don't make BEM structure too complicated. You don't need to create separate blocks for anything other than card itself and stars._

One last change to this PR would be to modify classes to have only 2 block elements: **card** and **stars**

DorotaLeniecDev (changes requested):
One small change and we are good to go :)

PR #2734, opened 2023-03-10
add task solution v1

SsebastianPawlik:
- [DEMO LINK](https://SsebastianPawlik.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://SsebastianPawlik.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/card.css
+++ b/src/card.css
@@ -1,0 +1,91 @@
+.card-container__product-card {
+  padding: 32px 16px 16px;
+}
+
+.product-card__title {
+  margin: 0 0 4px;
+  font-size: 12px;
+  line-height: 18px;
+  font-weight: 500;
+}
+
+.product-card {
+  display: flex;
+  flex-direction: column;
+  max-width: 200px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  align-items: center;
+  color: #060b35;
+}
+
+.product-card__image {
+  margin: 0 0 40px;
+  width: 160px;
+  height: 134px;
+}
+
+.product-card__code {
+  margin: 0 0 16px;
+  color: #616070;
+  font-size: 10px;
+  align-self: start;
+  line-height: 14px;
+}
+
+.product-card__reviews-text {
+  margin: 0;
+  font-size: 10px;
+}
+
+.product-card__reviews {
+  margin: 0 0 24px;
+  display: flex;
+  justify-content: space-between;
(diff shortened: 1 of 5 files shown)


danielpopek94 (approved):

src/card.css
```diff
@@ -0,0 +1,91 @@
...
+  font-weight: 700;
+  border-radius: 5px;
+  border: solid 1px transparent;
+  transition: 0.3s ease-in-out;
+}
+
+.product-card__buy-button:hover {
+  background-color: #fff;
+  color: #00acdc;
+  border-color: #00acdc;
```
d-wasilewski, line 90: Use variables for values that occur often

src/card.css
```diff
@@ -0,0 +1,91 @@
...
+  margin: 0;
+  font-weight: 700;
+}
+
+.product-card__buy-button {
+  background-color: #00acdc;
+  color: #fff;
+  width: 100%;
+  text-align: center;
+  min-height: 40px;
```
d-wasilewski, line 74: Use padding instead of height for styling buttons

src/card.css
```diff
@@ -0,0 +1,91 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  text-decoration: none;
+  text-transform: uppercase;
+  font-size: 14px;
+  font-weight: 700;
+  border-radius: 5px;
+  border: solid 1px transparent;
+  transition: 0.3s ease-in-out;
```
d-wasilewski, line 84:
```suggestion
  transition: <explicitly-specify-property> 0.3s ease-in-out;
```

d-wasilewski (changes requested):

src/card.css
d-wasilewski: It still persists. I meant something like `transition: color 0.3s ease-in-out;`. Write out all of them

src/card.css
```diff
@@ -0,0 +1,97 @@
+:root {
+  --color-aqua: #00acdc;
+  --color-white: #fff;
+  --color-grey: #616070;
+}
```
d-wasilewski, line 5: That's good! A better approach is to keep it in a separate file or style.css, so please move it

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2750, opened 2023-03-13
add task solution

SuaBy514:
- [DEMO LINK](https://suaby514.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://suaby514.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,74 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <title>Product cards</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Product cards</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
+    <title>Product cards</title>
+    <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="anonymous"
+    >
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+  </head>
+  <body>
+    <div class="card" data-qa="card">
+      <div class="card__imac">
+        <img
+          src="images/imac.jpeg"
+          alt="iMac A1419 MNED2UA/A"
+          class="card__imac card__imac--image"
+        >
+      </div>
+
+      <div class="card__product">
+        <p class="card__text card__text--title">
+          APPLE A1419 iMac 27" Retina<br>5K Monoblock (MNED2UA/A)
+        </p>
+      </div>
+
+      <div class="card__code">
+        <p class="card__text card__text--sub">
+          Product code: 195434
+        </p>
+      </div>
+
+      <div class="card__reviews">
+        <ul class="stars stars--4">
+          <li class="stars__star"></li>
+
+          <li class="stars__star"></li>
+
+          <li class="stars__star"></li>
+
+          <li class="stars__star"></li>
+
+          <li class="stars__star"></li>
+        </ul>
+
+        <p class="card__text card__text--review">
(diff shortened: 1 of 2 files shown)


Amsko6 (approved):
All seems to be working just fine :)

src/index.html
```diff
@@ -1,13 +1,74 @@
...
+        <p class="card__text card__text--price--title">
+          Price:
+        </p>
+
+        <p class="card__text card__text--price--value">
+          $2,199
+        </p>
+      </div>
+
+      <button class="card__button" data-qa="hover">
```
d-wasilewski, line 69: it's missing a `href` attribute

d-wasilewski (changes requested):

src/style.css
```diff
@@ -0,0 +1,148 @@
...
+  color: #060b35;
+  font-style: normal;
+  text-align: right;
+}
+
+.card__button {
+  margin-top: 4px;
+  padding: 0;
+  width: 166px;
+  height: 40px;
```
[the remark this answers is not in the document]
SuaBy514: Height is there because the button was, by default, bigger than it needed to be, it's to scale it down rather than up

src/style.css
```diff
@@ -0,0 +1,148 @@
...
+.stars--5 :nth-child(-n+5) {
+  background-image: url("images/star-active.svg");
+}
+
+/* Hover function */
+
+.card__button:hover {
+  background: transparent;
+  color: #00acdc;
+  border: 1px solid #00acdc;
```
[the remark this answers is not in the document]
SuaBy514: It refused to appear without redefining the border altogether, possibly because it's defined as simply transparent earlier in the file

src/style.css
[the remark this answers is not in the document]
SuaBy514: I figured it out, thank you

src/style.css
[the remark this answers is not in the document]
SuaBy514: After some attempts, I couldn't get it to work by just changing the border-color so long as border was defined as transparent in the non-:hover block

src/css/card.css
```diff
@@ -0,0 +1,121 @@
...
+  font-weight: var(--fwHigh);
+  color: #060b35;
+  font-style: normal;
+  text-align: right;
+}
+
+.card__button {
+  margin-top: 4px;
+  padding: 12px 0;
+  width: 166px;
```
d-wasilewski, line 99: width should be set to 100% here, it shouldnt be fixed

src/style.css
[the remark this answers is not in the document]
d-wasilewski: By defining it as `border: transparent` you don't define it at all (it's of size 0).
You need to set it explicitly like `border: 1px solid transparent`, and then you can change the color only
Please try that

src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  --textColorSecondary: #616070;
+  --buttonCyan: #00acdc;
+  --lhSmall: 14px;
+  --lhBig: 18px;
+  --fsSmall: 10px;
+  --fsBig: 12px;
+  --fwLow: 400;
+  --fwMid: 500;
+  --fwHigh: 700;
+}
```
d-wasilewski, line 16: It's duplicated (exists in card.css too). This place is good for this, but you can also put it in a seperate file which would be a better approach

src/css/card.css
```diff
@@ -0,0 +1,121 @@
...
+  color: var(--textColorSecondary);
+  font-style: normal;
+}
+
+.card__reviews,
+.card__price {
+  margin: 12px 0;
+  display: flex;
+  flex-direction: row;
+  width: 166px;
```
d-wasilewski, line 63: use percentage value

d-wasilewski (changes requested):

src/style.css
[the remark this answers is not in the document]
SuaBy514: Thank you, I will try that, and if it works I'll also apply it to the catalog

src/css/card.css
SuaBy514: Not defining the width at all also seems to work exactly the same way; it goes to 100% by default

src/style.css
SuaBy514: I will try to do that, I couldn't use variables from other files, that's why that was there... I will try to put it in another file

src/css/card.css
SuaBy514: Not defining it at all has the same effect as well

src/style.css
SuaBy514: For some reason moving it to a separate file seems to have broken the stars... but on the bright side, using the variables from just inside style.css works fine now

d-wasilewski (approved):

PR #2752, opened 2023-03-13
Add solution

mateusz-michalowski:
[DEMO LINK](https://mateusz-michalowski.github.io/layout_product-cards/)
[TEST REPORT LINK](https://mateusz-michalowski.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,63 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <title>Product cards</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Product cards</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content=
+          "width=device-width,
+           user-scalable=no,
+           initial-scale=1.0,
+           maximum-scale=1.0,
+           minimum-scale=1.0">
+    <title>Product cards</title>
+    <link
+    href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
+    rel="stylesheet"
+    >
+    <link rel="stylesheet" href="./style.css">
+  </head>
+  <body>
+    <div class="card" data-qa="card">
+      <img
+        src="./images/imac.jpeg"
+        alt="APPLE A1419 iMac 27"
+        class="card__product-image"
+      >
+      <p class="card__product-name">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </p>
+      <p class="card__product-code">
+        Product code: 195434
+      </p>
+      <div class="card___stars-box">
+        <div class="stars stars--4">
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star stars_last"></div>
+        </div>
+        <div class="reviews">
+          Reviews: 5
+        </div>
+      </div>
+      <div class="card__price">
+        <div class="card__value-text">
+          Price:
+        </div>
+        <div class="card__price-number">
+          $2,199
+        </div>
+      </div>
+      <div class="card__button">
+        <a
+          class="card__button-link"
+          href="#"
+          data-qa="hover"
+        >
+          buy
+        </a>
+      </div>
+    </div>
+  </body>
 </html>
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Product cards
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_product-cards/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_product-cards/report/html_report/)
+- [DEMO LINK](https://mateusz-michalowski.github.io/layout_product-cards/)
+- [TEST REPORT LINK](https://mateusz-michalowski.github.io/layout_product-cards/report/html_report/)
 
 > Follow [this instruction](https://mate-academy.github.io/layout_task-guideline)
 ___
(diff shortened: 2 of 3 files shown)


bjeziorek (changes requested):

bjeziorek (changes requested):
Don't hardcode exact height size - checklist point 8 - you still have in your css hardcoded height (card.css lines 62, 73). Instead you can use line-height and/or padding. Exception is the image - it's not even and there height is necessary, in all other places of card.css it should be removed. Rest seems fine :).

bjeziorek (approved):
Looks fine :).

src/card.css
```diff
@@ -0,0 +1,106 @@
...
+  font-weight: 400;
+  font-size: 10px;
+  line-height: 14px;
+  color: #060b35;
+  align-self: center;
+}
+
+.card__price {
+  display: flex;
+  width: 168px;
```
d-wasilewski, line 54: This shouldn't be hardcoded

src/card.css
```diff
@@ -0,0 +1,106 @@
...
+  margin-top: 40px;
+}
+
+.card__product-code {
+  margin-top: 4px;
+  font-style: normal;
+  font-weight: 400;
+  font-size: 10px;
+  line-height: 14px;
+  color: #616070;
```
d-wasilewski, line 32: Create variables for values that occur often

src/card.css
```diff
@@ -0,0 +1,106 @@
...
+
+.card___stars-box {
+  display: flex;
+  justify-content: space-between;
+  align-items: flex-end;
+}
+
+.card__reviews {
+  margin-top: 1px;
+  text-align: right;
```
d-wasilewski, line 44: This is redundant

src/card.css
```diff
@@ -0,0 +1,106 @@
...
+  font-weight: 400;
+  font-size: 10px;
+  line-height: 14px;
+  color: #060b35;
+  align-self: center;
```
d-wasilewski, line 49: Probably unnecessary too

src/card.css
```diff
@@ -0,0 +1,106 @@
...
+  text-decoration: none;
+}
+
+.card__button:hover {
+  background: white;
+}
+
+.card__button-link:hover {
+  color: #00acdc;
+}
```
d-wasilewski, line 106: You can join it

src/card.css
```diff
@@ -0,0 +1,106 @@
...
+  text-transform: uppercase;
+  text-align: center;
+  border: 1px solid #00acdc;
+}
+
+.card__button-link {
+  font-style: normal;
+  font-weight: 700;
+  font-size: 14px;
+  line-height: 38px;
```
d-wasilewski, line 94: You should style buttons with `padding` instead of `height`

src/card.css
```diff
@@ -0,0 +1,106 @@
...
+  font-size: 16px;
+  line-height: 18px;
+}
+
+.card__button {
+  box-sizing: border-box;
+  margin-top: 16px;
+  background: #00acdc;
+  border-radius: 5px;
+  width: 166px;
```
d-wasilewski, line 83: use percentage value

d-wasilewski (changes requested):
Apply suggestions in all places

src/card.css
```diff
@@ -0,0 +1,102 @@
+html {
+  --mainColor: #00acdc;
+}
```
d-wasilewski, line 3: Here's an example of how you should declare variables in CSS:

```
:root {
  --primary-color: #007bff;
  --secondary-color: #6c757d;
}

File: src/card.css [hunk 9]
```diff
@@ -0,0 +1,102 @@
...
+
+.card__price {
+  display: flex;
+  padding-top: 24px;
+}
+
+.card__value-text {
+  text-align: left;
+  flex-grow: 1;
```
d-wasilewski, line 61: You dont need it here and in price-number. Use flexbox on parent element to style them

src/index.html
```diff
@@ -1,13 +1,74 @@
...
+
+      <div class="card__button">
+        <a
+          class="card__button-link"
+          href="#"
+          data-qa="hover"
+        >
+          buy
+        </a>
+      </div>
```
d-wasilewski, line 71: You can have only <a> tag here, this <div> is not necessary and creates mess in css

d-wasilewski (changes requested):
Please fix previous suggestions aswell

d-wasilewski (changes requested):

src/card.css
d-wasilewski: Again, this is redundant, please remove this

d-wasilewski: And this is not the only place, get rid of this text-align property

src/card.css
```diff
@@ -0,0 +1,98 @@
...
+  font-size: 16px;
+  line-height: 18px;
+}
+
+.card__button {
+  box-sizing: border-box;
+  margin-top: 16px;
+  background: var(--primary-color);
+  border-radius: 5px;
+  bottom: 17px;
```
d-wasilewski, line 81: what is this supposed to do? Remove it aswell

src/card.css
```diff
@@ -0,0 +1,98 @@
...
+  text-transform: uppercase;
+  text-align: center;
+  border: 1px solid;
+  font-style: normal;
```
d-wasilewski, line 85: unnnecessary

src/card.css
```diff
@@ -0,0 +1,98 @@
...
+  font-weight: 700;
+  font-size: 14px;
+  padding: 12px 0;
+  color: white;
+  display: block;
```
d-wasilewski, line 90: default value

d-wasilewski (changes requested):

src/card.css
d-wasilewski: By saying it's default I meant you can remove this aswell. It's the last thing :)

d-wasilewski (changes requested):

src/card.css
mateusz-michalowski: I don't know which of these four lines is default, but removing any of these destroys button completly. Font size and padding give the 'shape' of the button, color changes default black, and becouse my button is not in div, but its modified 'a', default display is not block either.

d-wasilewski: You're right, mb

d-wasilewski (approved):

PR #2754, opened 2023-03-13
card solution

bjeziorek:
- [DEMO LINK](https://bjeziorek.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://bjeziorek.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/card.css
+++ b/src/card.css
@@ -1,0 +1,124 @@
+.card {
+  border: 1px solid rgba(243, 243, 243, 1);
+  border-radius: 5px;
+
+  font-family: "Roboto", Arial, Helvetica, sans-serif;
+
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+
+  width: 200px;
+}
+
+.card__photo {
+  width: 160px;
+  height: 134px;
+  margin: 32px 19px 0;
+}
+
+.card__name {
+  width: auto;
+  margin: 40px 16px 0;
+
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 18px;
+  color: var(--almost-black);
+}
+
+.card__code {
+  width: auto;
+  margin: 4px 16px 0;
+
+  font-weight: 400;
+  font-size: 10px;
+  line-height: 14px;
+  color: var(--warm-grey);
+}
+
+.card__rate {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: flex-end;
+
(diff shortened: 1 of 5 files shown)


mwgreybeard72 (approved):
Seems it all works just fine. However, I wouldn't have left any spaces between </head> and <body> Line: 14-16 <a> and </section> Line: 46-48 in your index.html file.

src/card.css
```diff
@@ -0,0 +1,124 @@
...
+}
+
+.card__photo {
+  width: 160px;
+  height: 134px;
+  margin: 32px 19px 0;
+}
+
+.card__name {
+  width: auto;
```
d-wasilewski, line 21: auto is default width's property, so you don't need to use it like that, it has no effect. If you need to use width somewhere you can try with % values, however I think deleting this won't have any effect

src/stars.css
```diff
@@ -0,0 +1,48 @@
...
+  background-image: url(images/star.svg);
+}
+  
+.stars--4 > div:nth-child(n + 5) {
+  background-image: url(images/star.svg);
+}
+  
+.stars--5 > div:nth-child(n + 6) {
+  background-image: url(images/star.svg);
+}
```
d-wasilewski, line 48: I left a comment in stars task already, but don't forget to change this here aswell once you're done with that

src/card.css
```diff
@@ -0,0 +1,124 @@
...
+  font-size: 14px;
+  line-height: 13px;
+  color: var(--white);
+
+  align-items: center;
+  padding: 12px;
+  width: auto;
+  margin: 16px;
+
+  box-sizing: border-box;
```
d-wasilewski, line 105: you've set it for each element in style.css already

src/card.css
```diff
@@ -0,0 +1,124 @@
...
+  text-align: center;
+  font-size: 14px;
+  color: var(--light-blue);
+
+  align-items: center;
+  background-color: var(--white);
+  border: 1px solid var(--light-blue);
+
+  cursor: pointer;
+}
```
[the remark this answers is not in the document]
d-wasilewski: Also, `border` should already be defined, but should be initially transparent. On hover you just want to change its color with `border-color` property

d-wasilewski (changes requested):

d-wasilewski (changes requested):
^

d-wasilewski (approved):

PR #2788, opened 2023-03-21
add task solution

LiliiaDanylenko:
- [DEMO LINK](https://LiliiaDanylenko.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://LiliiaDanylenko.github.io/layout_product-cards/report/html_report/)

Не зовсім розібралась із margin...

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -10,25 +10,28 @@
 
 .page-cards__content {
   margin: 0;
-  box-sizing: border-box;
 }
 /* #endregion */
 
 .card {
-  width: 198px;
+  box-sizing: border-box;
+  width: 200px;
   border: 1px solid rgb(243, 243, 243);
   border-radius: 5px;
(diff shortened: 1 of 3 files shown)


VitaliyBondarenko1982 (changes requested):

src/style.css
```diff
@@ -0,0 +1,133 @@
...
+  font-size: 14px;
+  line-height: 16px;
+  text-decoration: none;
+  color: rgb(255, 255, 255);
+
+  background: var(--color-blue);
+  border: 1px solid var(--color-blue);
+  border-radius: 5px;
+
+  width: 166px;
```
YegorKorenkov, line 123: if the element should be full width, then it is better to set width 100%

YegorKorenkov (approved):
Nice work, please check comment

PR #2792, opened 2023-03-22
add product card

MartaChobaniuk:
- [DEMO LINK](https://martachobaniuk.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://martachobaniuk.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -49,9 +49,9 @@
 
       <div class="product-card__button-container">
         <a
-        href="#"
-        class="product-card__button"
-        data-qa="hover"
+          href="#"
+          class="product-card__button"
+          data-qa="hover"
         >Buy
         </a>
       </div>
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,60 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <title>Product cards</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Product cards</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -1,13 +1,60 @@
...
+        <p class="product-card__price-value">$2,199</p>
+      </div>
+
+      <div class="product-card__button-container">
+        <a
+        href="#"
+        class="product-card__button"
+        data-qa="hover"
+        >Buy
+        </a>
```
Volodymyr-Melnyk-Fron10der, line 56:
```suggestion
        <a
          href="#"
          class="product-card__button"
          data-qa="hover"
        >
          Buy
        </a>
```

src/style.css
```diff
@@ -0,0 +1,130 @@
...
+p {
+  margin: 0;
+}
+
+.product-card {
+  width: 200px;
+  display: flex;
+  flex-direction: column;
+  background: var(--color-white);
+  border: 1px solid #f3f3f3;
```
Volodymyr-Melnyk-Fron10der, line 24: so move all color if you do it for some of them

src/style.css
```diff
@@ -0,0 +1,130 @@
...
+  border-radius: 5px;
+  padding: 32px 16px 16px;
+  font-family: "Roboto", Arial, Helvetica, sans-serif;
```
Volodymyr-Melnyk-Fron10der, line 27: move font family into html selector

src/style.css
```diff
@@ -0,0 +1,130 @@
...
+  font-style: normal;
```
Volodymyr-Melnyk-Fron10der, line 28: remove redundant default properties

Volodymyr-Melnyk-Fron10der (changes requested):
Hi! Good job👍

1. Check backstop report, there are some inconsistences

loralevitska (changes requested):
Almost done) Let's fix tests. Try to use line-height for `product-card__rewiews`
Then switch between two tabs 
image
Test - your solution
Reference - example

Check difference in tests and fix it
If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

vlad-demchuk (approved):
we don't require pixel perfect, however you could try to fix test with this styles for `product-card__rewiews`

```
line-height: 14px;
display: flex;
align-items: center;

PR #2798, opened 2023-03-24
Product card

GlibVK:
https://glibvk.github.io/layout_product-cards/
https://glibvk.github.io/layout_product-cards/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,14 +1,12 @@
-/* #region reset */
 body {
   margin: 0;
+  font-family: Roboto, sans-serif;
 }
 
 ul {
   margin: 0;
 }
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,151 @@
...
+  display: flex;
+  flex-direction: column;
+}
+
+.stars {
+  padding: 0;
+  display: flex;
+}
+
+.stars--4 li:nth-child(-n + 4) {
```
OleziO, line 149: Add code for all cases (from 1 to 5 stars)

OleziO (changes requested):

PR #2801, opened 2023-03-24
Product Card

GlibVK:
I have a question. Why when i create @import "./styles/star.css" - import don`t work, but when i write root @import "./stras.css" - its working?

https://glibvk.github.io/layout_product-cards/
https://glibvk.github.io/layout_product-cards/report/html_report/


src/reset.css
```diff
@@ -0,0 +1,13 @@
+* {
```
witflash, line 1: I think this file is not a "reset" exactly but something like "base" maybe. But ok :)

src/star.css
```diff
@@ -0,0 +1,25 @@
...
+  background-position: center;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
+
+.stars {
+  display: flex;
+}
```
witflash, line 17: It's better to place Block selector at the top of the file. And all Elements should be following after it.

src/styles/card.css
```diff
@@ -0,0 +1,108 @@
...
+  color: var(--color-dark-blue);
+}
+
+.card__id-text {
+  font-weight: 400;
+  font-size: var(--size-text-s);
+  line-height: 14px;
+  color: var(--color-gray);
+
+  margin-bottom: var(--min-element-margin);
```
witflash, line 44: I see margin rules somewhere placed at the top of the selector rules and somewhere at the bottom. Try to make your own approach with the same arrangement of the CSS rules through all files.

witflash (approved):
Well done 👍 

I left several comments to improve your future work.

PR #2810, opened 2023-03-28
add product cards

tniklevych:
- [DEMO LINK](https://tniklevych.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://tniklevych.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,135 @@
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap");
+
+:root {
+  --main-font-color: #060b35;
+  --secondary-font-color: #616070;
+  --blue-elements-color: #00acdc;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: Roboto, sans-serif;
+  font-style: normal;
+  font-weight: 400;
+  font-size: 10px;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,135 @@
...
+}
+
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: Roboto, sans-serif;
+  font-style: normal;
+  font-weight: 400;
```
sTorba24, line 16: it's default values, so you can remove it

src/style.css
```diff
@@ -0,0 +1,135 @@
...
+  font-size: 10px;
```
sTorba24, line 17: not best practice generally to set font-size for html, would be better to set it on element directly

src/style.css
```diff
@@ -0,0 +1,135 @@
...
+
+.page {
+  display: flex;
+}
+
+.card {
+  display: flex;
+  flex-direction: column;
+  width: 200px;
+  height: 406px;
```
sTorba24, line 32: don't set height manually, it can be done with content inside

sTorba24 (approved):
It's good! Some improvements need to be done :)

slavanosatskyi (approved):

PR #2813, opened 2023-03-29
Product card 5.0

Dimoshua:
- [DEMO LINK](https://Dimoshua.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://Dimoshua.github.io/layout_product-cards/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,91 @@
...
+  font-size: 10px;
+  line-height: 14px;
+  color: var(--text-second-color);
+}
+
+body {
+  margin: 0;
+}
+
+.card {
```
Volodymyr-Melnyk-Fron10der, line 16: move into separated file

src/style.css
```diff
@@ -0,0 +1,91 @@
...
+html {
+  font-family: Roboto, sans-serif;
+  font-size: 10px;
+  line-height: 14px;
+  color: var(--text-second-color);
+}
+
+body {
+  margin: 0;
+}
```
Volodymyr-Melnyk-Fron10der, line 14: you can move it into normalize.css

src/style.css
```diff
@@ -0,0 +1,91 @@
+@import "./blocks/variables.css";
+@import "./blocks/normalize.css";
+@import "./blocks/stars.css";
+
+html {
+  font-family: Roboto, sans-serif;
+  font-size: 10px;
+  line-height: 14px;
```
Volodymyr-Melnyk-Fron10der, line 8: it's not a common rules, remove it from the html selector and set only where you really need it

src/style.css
```diff
@@ -0,0 +1,91 @@
...
+}
+
+.card__price-value {
+  font-size: 16px;
+  line-height: 18px;
+  font-weight: bold;
+  color: var(--text-color);
+}
+
+.card__buy {
```
Volodymyr-Melnyk-Fron10der, line 72: add some transition to the button

src/style.css
```diff
@@ -0,0 +1,91 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 40px;
```
Volodymyr-Melnyk-Fron10der, line 76: do not hardcode height, use paddings

Volodymyr-Melnyk-Fron10der (changes requested):
Almost there!

src/style.css
Dimoshua: for example: transition: 2s;
???

PR #2845, opened 2023-04-05
add task solution

mrsargos:
- [DEMO LINK](https://mrsargos.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://mrsargos.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/styles/index.css
+++ b/src/styles/index.css
@@ -1,0 +1,16 @@
+@import "./card.css";
+@import "./stars.css";
+
+* {
+  box-sizing: border-box;
+  font-family: "Roboto", sans-serif;
+  --color: #00adcd;
+  --color-text: #060b35;
+  --color-text-code-price: #616070;
+}
+
+body,
+h1,
+p {
+  margin: 0;
+}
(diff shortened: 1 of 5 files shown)


src/styles/index.css
```diff
@@ -0,0 +1,16 @@
+@import "./card.css";
+@import "./stars.css";
+
+* {
+  box-sizing: border-box;
+  font-family: "Roboto", sans-serif;
+  --color: #00adcd;
```
OleziO, line 7: Move to the :root selector.

OleziO (changes requested):

src/styles/index.css
```diff
@@ -0,0 +1,19 @@
...
+  box-sizing: border-box;
+}
+
+:root {
+  --color: #00adcd;
+  --color-text: #060b35;
+  --color-text-code-price: #616070;
+  font-family: "Roboto", sans-serif;
+}
+
```
IvanFesenko, line 14: `font-family` should be in `html` selector

IvanFesenko (approved):
Check my comment to avoid mistakes in the future

src/styles/index.css
mrsargos: Understood, thank you

PR #2848, opened 2023-04-06
add task solution

VolodymyrPst:


src/index.html
```diff
@@ -1,13 +1,52 @@
...
+           Price: <span>$2,199</span>
+          </div>
+        </div>
+
+        <button
+          class="card__button"
+          data-qa="hover"
+        >
+          BUY
+        </button>
```
anstsot, line 48: buy button should be a link. please use `a` tag

src/style.css
```diff
@@ -0,0 +1,123 @@
...
+
+.card__image {
+  width: 160px;
+  height: 134px;
+  margin-bottom: 40px;
+}
+
+/* .card__info {
+
+} */
```
anstsot, line 38: remove comments please

anstsot (changes requested):
Good work but there are some missing points. Please pay attention to these:
- Reuse stars block from [Stars task](https://github.com/mate-academy/layout_stars) and keep it operational. If you change stars--4 modifier to stars--2 or other - number of selected stars should change correctly.
- There should be only 2 BEM blocks in your code card and stars all the others are just card elements

PR #2876, opened 2023-04-14
Product cards

anpiet:
[DEMO LINK](https://anpiet.github.io/layout_product-cards/)
[TEST REPORT LINK](https://anpiet.github.io/layout_product-cards/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,189 @@
+/* #region Reset */
+:root {
+  box-sizing: border-box;
```
asakevych, line 3: it's a good practice to set box-sithing for all elements - `* { box-sizing: border-box; }`

src/style.css
```diff
@@ -0,0 +1,189 @@
...
+  margin: 0;
+}
+/* #endregion */
+/* #region Card */
+.card {
+  position: relative;
+  left: 0;
+  top: 0;
+  width: 198px;
+  height: 406px;
```
asakevych, line 25: 1. position is unnecessary here
2. you can use display flex with flex-direction column 
3. width should have 200px with border, so just set it too 200px (check box-sizing once more)
4. it's better not to set height here, let's make content to dedicate it

src/index.html
```diff
@@ -1,13 +1,50 @@
...
+  <body>
+    <div class="card" data-qa="card">
+      <div class="card__product-photo">
+      </div>
+      <div class="card__product-title">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </div>
+
+      <div class="card__product-code">
+        Product code: 195434
```
asakevych, line 22: it's unnecessary to add everywhere prefix 'product-', you can just call 'card__image', 'card__title' - let's simplify those names

src/index.html
```diff
@@ -1,13 +1,50 @@
...
+      <div class="stars stars--4 card__stars">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="card__review">
+          Reviews: 5
+      </div>
```
asakevych, line 33: please add some container for those elements 
```suggestion
<div class="card__rating">
      <div class="card__stars stars stars--4">
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
      </div>
      <div class="card__reviews">
          Reviews: 5
      </div>
</div>
```

src/index.html
```diff
@@ -1,13 +1,50 @@
...
+      <div class="card__price">
+        Price:
+      </div>
+      <div class="card__amount">
+        $2,199
+      </div>
```
asakevych, line 39: those elements also should have some container, it would be easier to pose them appropriatly

src/style.css
```diff
@@ -0,0 +1,189 @@
...
+}
+
+.card__product-title {
+  text-decoration: none;
+  position: absolute;
+  width: 166px;
+  left: 16px;
+  top: 206px;
+  color: #060b35;
+}
```
asakevych, line 48: it's a good practice to avoid position absolute, here you can just use margins to set suitable spaces for all elements
https://dev.to/domagojvidovic/position-absolute-and-relative-the-simplest-guide-5169

When To Use It
If you need to adjust the layout within a page or a component, you should avoid position: absolute; in 99% of the cases. Opt-in for Flexbox or Grid instead.

However, if you want to add a badge to your image, it's just perfect. You always want the badge to be on the top (or the edge) of the image, so that it seems like it's a part of an image.

Or if you want to stack a few images/elements onto each other. There's no way other than position: absolute;.

And that's the ultimate guide: use absolute positioning only when there's no other way to achieve that.

src/style.css
```diff
@@ -0,0 +1,189 @@
...
+
+.card__product-code {
+  position: relative;
+  width: 98px;
+  height: 14px;
+  top: 246px;
+  left: 16px;
+
```
asakevych, line 56: 1. position is also unnecessary, use margins
2. deafault width value is fit-content here, so it will fit your text inside, so you can also remove width here

src/style.css
```diff
@@ -0,0 +1,189 @@
...
+.card__review {
+  position: absolute;
+  left: 135px;
+  top: 276px;
+  height: 16px;
+  width: 53px;
+  font-weight: 400;
+  font-size: 10px;
+  color: var(--Main-accent);
+}
```
asakevych, line 80: the same with poisiton, make this element inside one conteiner with stars

src/style.css
```diff
@@ -0,0 +1,189 @@
...
+.card__price {
+  position: absolute;
+  left: 16px;
+  top: 316px;
+  width: 31px;
+  height: 18px;
+  color: var(--Secondary);
+  font-family: "Roboto", Arial, Helvetica, sans-serif;
+  font-weight: 400;
+}
```
asakevych, line 91: 1. same here with position, try to use container and flex and also give it margins
2. font-family can be dedicated for the whole body

src/style.css
```diff
@@ -0,0 +1,189 @@
...
+  text-align: right;
+  left: 132px;
+  top: 316px;
+  width: 50px;
+  height: 18px;
+  font-family: "Roboto", Arial, Helvetica, sans-serif;
+  font-size: 16px;
+  font-weight: 700;
+  color: var(--Main-accent);
+}
```
asakevych, line 104: same with position absolute

src/style.css
```diff
@@ -0,0 +1,189 @@
...
+
+.card__button {
+  box-sizing: border-box;
+  position: absolute;
+  text-transform: uppercase;
+  height: 40px;
+  width: 166px;
+  left: 16px;
```
asakevych, line 112: same with position absolute, use margin to add spaces and do not use position here :)

src/style.css
```diff
@@ -0,0 +1,189 @@
...
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+
+.card__product-photo {
+  position: absolute;
+  left: 19px;
+  top: 32px;
+  width: 160px;
+  height: 134px;
```
asakevych, line 35: try to use margins insted of position absolute

asakevych (changes requested):
Good job, but still it can be better 😉

src/style.css
```diff
@@ -0,0 +1,144 @@
...
+  background-image: url("images/imac.jpeg");
+  background-position: center;
+  background-size: 160px 134px;
+}
+
+.card__title {
+  text-decoration: none;
+
```
asakevych, line 44: `text-decoration: none;` do nothing here, it's not needed

src/style.css
```diff
@@ -0,0 +1,144 @@
...
+  padding: 0 16px;
+}
+
+.card__image {
+  margin: 32px auto 40px;
+  width: 160px;
+  height: 134px;
+  background-image: url("images/imac.jpeg");
+  background-position: center;
+  background-size: 160px 134px;
```
asakevych, line 39: please use image with alt instead of background image, it's is required due to accessibility:
image is important thing for a product and it would be easier for screen readers to know that here is an image of the product, while background-image is not noticed by screen readers and search engines

asakevych (changes requested):
so far so good! please use scss insted of css, you can use nested selectors, they work great with BEM 😉 and please deploy your project, there are your old changes in link

src/style/blocks/stars.scss
```diff
@@ -0,0 +1,27 @@
...
+    background-position: center;
+    line-height: 16px;
+  }
+
+  &__star:not(:last-child) {
+    margin-right: 4px;
+  }
+
+  @for $i from 1 through 5 {
+    &--#{$i} > .stars__star:not(:nth-child(n + #{$i + 1})) {
```
JoyCoffeeAddict, line 19: Could you think about simplifying this loop?

src/style/utils/_variables.scss
```diff
@@ -0,0 +1,5 @@
+$Main-accent: #060b35;
```
JoyCoffeeAddict, line 1: Typically, we start variable names with lowercase.

JoyCoffeeAddict (changes requested):

src/index.html
```diff
@@ -1,13 +1,55 @@
...
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+        </div>
+        <div class="card__review">
+            Reviews: 5
+        </div>
+      </div>
+      <div class="price-box card__price-box">
```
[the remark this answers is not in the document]
anpiet: class price-box if for creating a block in which I have to position elements (price-box__price & price-box__amount). card__price-box is because it's element in card block and it needs to be positioned with this class.  Can you explain how can I do it different way without use of both classes?

src/style/blocks/cards.scss
```diff
@@ -0,0 +1,78 @@
...
+    justify-content: space-between;
+    margin-bottom: 16px;
+  }
+
+  &__button {
+    text-transform: uppercase;
+    height: 40px;
+    width: 166px;
+    left: 16px;
+    bottom: 16px;
```
asakevych, line 62: 1. for inline-elements you cant set height, it's unnecessary here
2. position is static - you don't need to set left and bottom

src/style/blocks/cards.scss
```diff
@@ -0,0 +1,78 @@
...
+    width: 160px;
+    height: 134px;
+    background-image: url("images/imac.jpeg");
+    background-position: center;
+    background-size: 160px 134px;
+  }
+
+  &__title {
+    text-decoration: none;
+
```
asakevych, line 24: 1. unnecessary empty line
2.  `text-decoration: none;` is a default value here, it's unnecessary

src/style/blocks/cards.scss
```diff
@@ -0,0 +1,78 @@
...
+  border-radius: 5px;
+  padding: 0 16px;
+
+  &__image {
+    margin: 32px auto 40px;
+    width: 160px;
+    height: 134px;
+    background-image: url("images/imac.jpeg");
+    background-position: center;
+    background-size: 160px 134px;
```
asakevych, line 19: please add image by using `` tag instead of background image and add descriptive alt attributte as is required due to the the checklist

src/style/blocks/cards.scss
```diff
@@ -0,0 +1,78 @@
+@import "../utils/variables";
+
+.card {
+  display: flex;
+  flex-direction: column;
+  align-items: stretch;
+  width: 200px;
+  height: 406px;
```
asakevych, line 8: do not set height here, it's better content to idicate it

asakevych (changes requested):
Good job, but some changes are still needed, please check once more checklist before requesting review once more 😉

asakevych (approved):
Good job!! 🚀 🔥 🔥 
I still want you to change alt for image (it is required due to accessibility - screen readers and serch engines use alt attribute) , but the rest is good :)

PR #2958, opened 2023-04-25
add task solution

Huracann:
- [DEMO LINK](https://Huracann.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://Huracann.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/styles/card.css
+++ b/src/styles/card.css
@@ -1,0 +1,92 @@
+.card {
+  width: 200px;
+  padding: 32px 16px 16px;
+  box-sizing: border-box;
+  background: #fff;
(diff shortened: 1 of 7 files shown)


src/styles/card.css
```diff
@@ -0,0 +1,92 @@
+.card {
+  width: 200px;
+  padding: 32px 16px 16px;
+  box-sizing: border-box;
```
Anna-Tkachenko, line 4: Will be better to set this property for all elements using `*` selector

Anna-Tkachenko (approved):
Good job! ⭐️

PR #2972, opened 2023-04-28
 sorry but I still can't understand what needs to be changed

SerhiiVizir:
- [DEMO LINK](https://SerhiiVizir.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://SerhiiVizir.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,70 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <title>Product cards</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Product cards</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <title>Product cards</title>
+    <link rel="stylesheet" href="./style.css">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+  </head>
+  <body>
+    <div class="card" data-qa="card">
+      <div class="card__logo">
+
+        <img
+          src="/images/imac.jpeg"
+          alt="imac-image"
+          class="card__logo-image"
+        >
+      </div>
+
+
+      <h2 class="card__title">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </h2>
+
+      <p class="card__code">
+        Product code: 195434
+      </p>
+
+      <div class="card__rating">
+        <div class="stars stars--4">
+          <div class="stars__star">
+          </div>
+          <div class="stars__star">
+          </div>
+          <div class="stars__star">
+          </div>
+          <div class="stars__star">
+          </div>
+          <div class="stars__star">
+          </div>
+        </div>
+
+        <p class="card__rating-reviw">
+          Reviews: 5
+        </p>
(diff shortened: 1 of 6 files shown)


src/styles/card.css
```diff
@@ -0,0 +1,95 @@
+.card {
+  display: flex;
```
witflash, line 2: Could you remove flex-box for the card block?
There are several blocks (they will be arranged vertically by default) inside it and there is no need to use flex here. It may cause some problem in your tests.

witflash (changes requested):
Could your check my comments please? And try to fix the tests.

SerhiiVizir:
it's better but still something is wrong

TarasHoliuk:
About the mismatch with the mockup:

in the test easy to see that your card has a bigger height and lower width than in the reference:
image

also, check all margins/paddings and font-weight
for example, there are too big vertical margins around `Price: $2,199`

src/index.html
```diff
@@ -1,13 +1,70 @@
...
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <title>Product cards</title>
+    <link rel="stylesheet" href="./style.css">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+  </head>
+  <body>
+    <div class="card" data-qa="card">
+      <div class="card__logo">
+
```
TarasHoliuk, line 14: redundant line between parent and child tags

src/index.html
```diff
@@ -1,13 +1,70 @@
...
+        <img
+          src="/images/imac.jpeg"
+          alt="imac-image"
+          class="card__logo-image"
+        >
+      </div>
+
+
```
TarasHoliuk, line 22: only one line is allowed between sibling elements

src/styles/card.css
```diff
@@ -0,0 +1,93 @@
+.card {
+  flex-direction: column;
+  box-sizing: border-box;
```
TarasHoliuk, line 3: set it in `*` selector

src/styles/card.css
```diff
@@ -0,0 +1,93 @@
...
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  width: 200px;
+  padding: 32px 16px 16px;
+}
+
+.card__logo {
+  justify-content: center;
```
TarasHoliuk, line 11: redundant (doesn't work without `display: flex`), so just delete it

src/styles/reset.css
```diff
@@ -0,0 +1,10 @@
+html {
+  font-family: Roboto, sans-serif;
+
+  --main-color: #060b35;
+  --secondary-color: #616070;
+}
+
+body {
+  margin: 0;
```
TarasHoliuk, line 9: reset margin for `p` as well here

src/styles/reset.css
```diff
@@ -0,0 +1,10 @@
+html {
+  font-family: Roboto, sans-serif;
+
+  --main-color: #060b35;
+  --secondary-color: #616070;
+}
+
+body {
```
TarasHoliuk, line 8:
```suggestion
body, p {
```

src/styles/card.css
```diff
@@ -0,0 +1,93 @@
+.card {
+  flex-direction: column;
```
TarasHoliuk, line 2: redundant without `flex`

TarasHoliuk (changes requested):

SerhiiVizir:
I can't find what's wrong, I see that margin  possible, but they  all the same as in the figma

Volodymyr-Melnyk-Fron10der (changes requested):
pls, check these paddings and if you will still struggle with it - ask for some help in Slack, not here.
And check the previous comments, some of them are still not fixed 
image

nazarmatsevych (approved):

PR #2993, opened 2023-05-05
add task solution

MariaSnegireva:
- [DEMO LINK](https://MariaSnegireva.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://MariaSnegireva.github.io/layout_product-cards/report/html_report/)


src/fonts.css
```diff
@@ -0,0 +1,17 @@
...
+
+@font-face {
+  font-family: Roboto-medium;
+  src: url(./fonts/Roboto-Medium.ttf);
+}
+
+@font-face {
+  font-family: Roboto-bold;
+  src: url(./fonts/Roboto-Bold.ttf);
+}
```
Volodymyr-Melnyk-Fron10der, line 15: set font-weight for each face

Volodymyr-Melnyk-Fron10der (changes requested):

src/fonts.css
MariaSnegireva: done

vlad-demchuk (approved):

PR #3021, opened 2023-05-13
Final changes

Orchibald:
https://orchibald.github.io/layout_product-cards/
https://orchibald.github.io/layout_product-cards/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -2,20 +2,56 @@
 
 body {
   margin: 0;
-  font-family: "Roboto", serif;
+  font-family: "Roboto", sans-serif;
+  color: #060b35;
+
+  --secondary-color: #616070;
 }
 
 .card {
-  background: #fff;
-  border: #f3f3f3 1px solid ;
-  border-radius: 5px;
-  width: 200px;
-}
-
-.card__container {
-  margin: 16px;
   display: flex;
   flex-direction: column;
+  width: 200px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,116 @@
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap");
+
+body {
+  margin: 0;
+  font-family: "Roboto", sans-serif;
```
VitaliyBondarenko1982, line 5: font-family better to move to html selector

src/style.css
```diff
@@ -0,0 +1,116 @@
...
+  color: #060b35;
+
+  --secondary-color: #616070;
```
VitaliyBondarenko1982, line 8: variables better to create in :root selector

src/style.css
```diff
@@ -0,0 +1,116 @@
...
+  text-decoration: none;
+  background: #00acdc;
+  border-radius: 5px;
+  box-sizing: border-box;
+  border: 1px solid #00acdc;
+}
+
+.card__buy-item:hover {
+  background: #fff;
+  color: #00acdc;
```
VitaliyBondarenko1982, line 115: this color use more than one time
create variable for it

VitaliyBondarenko1982 (changes requested):
Good job.
Left a couple minor comments

PR #3023, opened 2023-05-13
Final task solution

dimkamg21:
https://dimkamg21.github.io/layout_product-cards/
https://dimkamg21.github.io/layout_product-cards/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,118 @@
+body {
+  margin: 0;
+  font-family: "Roboto", sans-serif;
+  color: #060b35;
+
+  --secondary-color: #616070;
+}
+
+.card {
+  display: flex;
+  flex-direction: column;
+  width: 200px;
+  box-sizing: border-box;
+  padding: 32px 16px 16px;
+
+  background: #fff;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+
+.card__img {
+  width: 160px;
+  height: 134px;
+  margin: 0 3px 40px;
+}
+
+.card__text {
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 18px;
+  margin: 0 0 4px;
+}
+
+.card__add-info,
+.card__rating-feedbacks {
+  font-weight: 400;
+  font-size: 10px;
+  line-height: 14px;
+  margin: 0;
+  /* vertical-align: middle; */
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,120 @@
+:root {
+  --secondary-color: #616070;
+  --buy-link-color: #00acdc;
+}
+
+body {
+  margin: 0;
+  font-family: "Roboto", sans-serif;
```
VitaliyBondarenko1982, line 8: font-family better move to html selector

src/style.css
```diff
@@ -0,0 +1,120 @@
...
+
+.card__add-info {
+  color: var(--secondary-color);
+  margin: 0 0 16px;
+}
+
+.card__rating {
+  display: flex;
+  align-items: center;
+  /* align-content: center; */
```
VitaliyBondarenko1982, line 53: remove commented code

src/style.css
```diff
@@ -0,0 +1,120 @@
...
+
+  justify-content: space-between;
+  height: 16px;
```
VitaliyBondarenko1982, line 56: looks like redundant height and it is bad practise to set height directly if it is not necessary

src/style.css
```diff
@@ -0,0 +1,120 @@
...
+}
+
+.card__price {
+  display: flex;
+  justify-content: space-between;
+  margin: 0 0 16px;
+}
+
+.card__price-text {
+  font-weight: 400;
```
VitaliyBondarenko1982, line 87: redundant. it is default font-weight

src/style.css
```diff
@@ -0,0 +1,120 @@
...
+.card__price-amount {
+  font-weight: 700;
+  font-size: 16px;
+  line-height: 18px;
+  margin: 0;
+}
+
+.card__buy-item {
+  width: 100%;
+  height: 40px;
```
VitaliyBondarenko1982, line 103: redundant. you already set line-height to 40px.
and proofread [checklist](https://github.com/mate-academy/layout_product-cards/blob/master/checklist.md)
Screenshot from 2023-05-13 17-46-42

set border for .card__buy-item and change only border-color on hover

VitaliyBondarenko1982 (changes requested):

PR #3029, opened 2023-05-15
add task solution

IhorVerba:
https://IhorVerba.github.io/layout_product-cards/
https://IhorVerba.github.io/layout_product-cards/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,117 @@
+:root {
+  --cardMainColor: #060b35;
+  --cardSecondarycolor: #616070;
+  --buttonColor: #00acdc;
+}
+
+.page {
+  margin: 0;
+}
+
+.card {
+  display: flex;
+  flex-direction: column;
+  width: 166px;
+  padding: 32px 16px 16px;
+  background-color: #fff;
+  font-family: "Roboto", helvetica, sans-serif;
+  border: 1px solid #f3f3f3 ;
+  border-radius: 5px;
+}
+
+.card__image {
+  width: 160px;
+  height: 134px;
+  object-position: center;
+  margin: 0 auto 40px;
+}
+
+.card__title {
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 18px;
+  color: var(--cardMainColor);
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,117 @@
...
+  object-position: center;
+  margin: 0 auto 40px;
+}
+
+.card__title {
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 18px;
+  color: var(--cardMainColor);
+  padding: 0 0 4px;
```
vlad-demchuk, line 34: you can reset padding for tag and just set padding for one side

vlad-demchuk: don't forget about css properties order
http://codeguide.academy/html-css.html#css-order

src/style.css
```diff
@@ -0,0 +1,117 @@
...
+}
+
+.card__code {
+  font-weight: 400;
```
vlad-demchuk, line 38: 400 is a default value

src/style.css
```diff
@@ -0,0 +1,117 @@
...
+
+.stars--1 .stars__star:first-child,
+.stars--2 .stars__star:nth-child(-n + 2),
+.stars--3 .stars__star:nth-child(-n + 3),
+.stars--4 .stars__star:nth-child(-n + 4),
+.stars--5 .stars__star:nth-child(-n + 5) {
+  background-image: url(images/star-active.svg);
+}
+
+.reviews {
```
vlad-demchuk, line 71: image
no need to create bem block here

vlad-demchuk: also, check this checklist point
image

src/style.css
```diff
@@ -0,0 +1,117 @@
...
+}
+
+.priceTitle {
+  font-weight: 400;
+  font-size: 12px;
+  line-height: 18px;
+  color: var(--cardSecondarycolor);
+}
+
+.priceAmount {
```
vlad-demchuk, line 92: redundant bem blocks

vlad-demchuk (changes requested):

PR #3057, opened 2023-05-20
add task solution

SuprunArtem:
[DEMO LINK](https://suprunartem.github.io/layout_product-cards/)
[TEST REPORT LINK](https://suprunartem.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,121 @@
+:root {
+  --color-dark: #060b35;
+  --color-lighter: #616070;
+  --color-background: #00acdc;
+}
+
+html {
+  font-family: "Roboto", sans-serif;
+}
+
+.body {
+  margin: 0;
+}
+
+.card {
+  display: flex;
+  flex-direction: column;
+  width: 200px;
+  box-sizing: border-box;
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -0,0 +1,121 @@
...
+
+html {
+  font-family: "Roboto", sans-serif;
+}
+
+.body {
+  margin: 0;
+}
+
+.card {
```
SanyaBratashchuk, line 15: Each BEM block should be in the separate file

src/style.css
```diff
@@ -0,0 +1,121 @@
...
+  background-position: center;
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars__star:nth-child(-n+4) {
+  background-image: url(images/star-active.svg);
+}
```
SanyaBratashchuk, line 72: reuse all styles from task with stars

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):
LGTM

PR #3074, opened 2023-05-23
Product card 7.0

Dimoshua:
description:
- [DEMO LINK](https://Dimoshua.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://Dimoshua.github.io/layout_product-cards/report/html_report/)


DarkMistyRoom (changes requested):
Good job! Just one fix:

Be consistent with your margins. Add only top or bottom, don't add both.
image
image

PR #3109, opened 2023-05-29
Card

Kacper-Leman:
https://Kacper-Leman.github.io/layout_product-cards/  - Demo Link
https://Kacper-Leman.github.io/layout_product-cards/report/html_report/  - Test Report Link

### DIFF
--- a/src/card.css
+++ b/src/card.css
@@ -1,0 +1,79 @@
+.card {
+  width: 200px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  padding: 32px 16px 16px;
+  background-color: #fff;
+}
+
+.card__img {
+  display: block;
+  margin: 0 auto 40px;
+  width: 160px;
+  height: 134px;
+
+}
+
+.card__text {
+  margin: 0 0 4px;
+  color: var(--text-color);
+  line-height: 18px;
+  font-weight: 500;
+  font-size: 12px;
+}
+
+.card__product {
+  margin-bottom: 16px;
+}
+
+.card__stars {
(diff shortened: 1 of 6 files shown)


src/card.css
```diff
@@ -0,0 +1,79 @@
...
+  line-height: 18px;
+  font-weight: bold;
+  color: var(--text-color);
+}
+
+.card__button {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  box-sizing: border-box;
```
Radoslaw-Czerniawski, line 63: You are already setting border-box on all elements, this is redundant

src/card.css
```diff
@@ -0,0 +1,79 @@
...
+.card__img {
+  display: block;
+  margin: 0 auto 40px;
+  width: 160px;
+  height: 134px;
+
+}
+
+.card__text {
+  margin: 0 0 4px;
```
Radoslaw-Czerniawski, line 18: Would be more descriptive to just use margin-bottom. If you need to reset margins, do it in * {...} rule

src/style.css
```diff
@@ -1,3 +1,18 @@
...
+}
+
+html {
+  font-family: Roboto, sans-serif;
+  font-size: 10px;
+  line-height: 14px;
+  color: var(--text-second-color);
+}
+
+p {
```
Radoslaw-Czerniawski, line 16: Don't style using tags (apart from unique non-repeated elements like html, body etc.)

Radoslaw-Czerniawski (changes requested):
Looks good, needs some changes code-wise

PR #3116, opened 2023-05-30
add task solution

Teerveet:


src/styles/index.scss
```diff
@@ -1,3 +1,128 @@
...
+
+  &__reviews {
+    margin: 0;
+
+    color: #060b35;
+    font-weight: 400;
+    font-size: 10px;
+    line-height: 14px;
+  }
+}
```
sTorba24, line 128: create separated file for each BEM block

sTorba24 (approved):
Approved!

Teerveet:
After separating the BEM block, the links stopped working, no matter how differently I wrote them. Also, `data-qa` does not find, although they are in index.html

PR #3180, opened 2023-06-03
add task solution

AnastasiiaKriuchkova:
- [DEMO LINK](https://<AnastasiiaKriuchkova>.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://<AnastasiiaKriuchkova>.github.io/layout_product-cards/report/html_report/)
- I can-t pass the test

### DIFF
--- a/src/styles/blocks/body.scss
+++ b/src/styles/blocks/body.scss
@@ -1,0 +1,12 @@
+html {
+  --color-text-dark: #060b35;
+  --color-text-light: #616070;
+}
+
+body,
+h1,
+p,
+a {
+  font-family: "Roboto", Helvetica, sans-serif;
+  margin: 0;
+}
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,65 @@
+.card {
+  box-sizing: border-box;
+  width: 200px;
+  background: #fff;
+  border: 1px solid #f3f3f3;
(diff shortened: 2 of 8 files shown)


src/styles/blocks/body.scss
```diff
@@ -0,0 +1,12 @@
+html {
+  --color-text-dark: #060b35;
+  --color-text-light: #616070;
```
OleziO, line 3: Move to the :root selector

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,65 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  box-sizing: border-box;
```
OleziO, line 4: Move to the * selector and remove from all other places

OleziO (changes requested):

src/styles/blocks/body.scss
AnastasiiaKriuchkova: done

src/styles/blocks/card.scss
AnastasiiaKriuchkova: if i move this to the * selector, my stars and price are become like column

etojeDenys (changes requested):

Volodymyr-Melnyk-Fron10der (approved):

PR #3197, opened 2023-06-10
layout Product card is done

vladkorobka:
https://vladkorobka.github.io/layout_product-cards/

### DIFF
--- a/src/styles/button.scss
+++ b/src/styles/button.scss
@@ -1,0 +1,35 @@
+.button {
+  border: none;
+  border-radius: 5px;
+  width: 100%;
+  height: 40px;
+  box-sizing: border-box;
+  transition: all 144ms;
+
+  &--buy {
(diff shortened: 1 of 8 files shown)


src/styles/button.scss
```diff
@@ -0,0 +1,35 @@
...
+    // text-align: center;
+    // text-transform: uppercase;
+  }
+
+  &--buy:hover {
+    border: 1px solid $button-border-color-hover;
+    background-color: $button-background-color-hover;
+    color: $button-font-color-hover;
+  }
+}
```
nazarmatsevych, line 35: No need to create separate file, and no need to add additional class, just use only one [card__button](url)
Also remove commented code

nazarmatsevych (changes requested):
Good work!
Pls check comments :)

nazarmatsevych (approved):

PR #3229, opened 2023-06-20
task solution

killfox93:
https://killfox93.github.io/layout_product-cards/
https://killfox93.github.io/layout_product-cards/report/html_report/


src/style.scss
```diff
@@ -0,0 +1,8 @@
+* {
+  margin: 0;
+  box-sizing: border-box;
+}
```
anastasiiavorobiova, line 4: Consider adding general stylings and resets to the separate file to keep your main styles file clean

agniya-i (approved):

PR #3250, opened 2023-06-29
First commit

MelnykA:
https://MelnykA.github.io/layout_product-cards/
https://MelnykA.github.io/layout_product-cards/report/html_report/

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Product cards
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_product-cards/)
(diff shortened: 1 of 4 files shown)


polosanya (changes requested):
Almost done!

src/styles/card.scss
```diff
@@ -0,0 +1,87 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  width: 200px;
```
polosanya, line 4: It would be great to create variables for all main values

polosanya (changes requested):
There is no need to store both css and scss files. You can delete this folder
image

polosanya (approved):
Well done!

PR #3259, opened 2023-07-05
task failure

domalewski:
[DEMO LINK](https://domalewski.github.io/layout_product-cards/)
[TEST REPORT LINK](https://domalewski.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -3,11 +3,46 @@
   <head>
     <meta charset="UTF-8">
     <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Product cards</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+    <link rel="stylesheet" href="./styles/index.scss">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card" >
+      <div class="card__foo">
+        <img
+        class="card__image"
+        src="images/imac.jpeg"
+        alt="Apple A1419 iMac 27&quot;"
+        >
+      </div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -3,11 +3,46 @@
...
-    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="./styles/index.scss">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card" >
+      <div class="card__foo">
```
Oskinavara, line 15: Foo is certainly not a good element name.

src/index.html
```diff
@@ -3,11 +3,46 @@
...
+        <img
+        class="card__image"
```
Oskinavara, line 17: Indentation is wrong here. Please use autoformatter

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,114 @@
...
+      font-family: $font-family;
+      font-weight: 700;
+      line-height: 18px;
+      color: $main-color;
+    }
+  }
+
+  &__button {
+    margin-top: 16px;
+    width: 164px;
```
Oskinavara, line 80: Setting width as fixed px value is not the best idea. Such button will not be responsive. It should have 100% width.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,114 @@
...
+    height: 38px;
```
Oskinavara, line 81: Height could be determined by the content + paddings. Setting fixed height is not optimal idea.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,114 @@
+@import "../utils/variables";
+
+.card {
+  width: 200px;
```
Oskinavara, line 4: Setting width as fixed is not the best idea most of the time. It would be better to set max-width and set width as 100%.

Oskinavara (changes requested):
Tests do not pass. There is something wrong with fonts.

src/styles/blocks/card.scss
domalewski: Yes, you are rigth but requirements says that card width is `200px` including border

choeqq (approved):
Looks good! Well done 🚀

PR #3265, opened 2023-07-05
task solution

Sancz0pansa:
[DEMO LINK](https://Sancz0pansa.github.io/layout_product-cards/)
[TEST REPORT LINK](https://Sancz0pansa.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Product cards
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_product-cards/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_product-cards/report/html_report/)
+- [DEMO LINK](https://Sancz0pansa.github.io/layout_product-cards/)
+- [TEST REPORT LINK](https://Sancz0pansa.github.io/layout_product-cards/report/html_report/)
 
 > Follow [this instruction](https://mate-academy.github.io/layout_task-guideline)
 ___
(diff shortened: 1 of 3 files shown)


choeqq (commented):
Were you propmt to use SCSS our CSS solution is also OK?

Sancz0pansa:
> Were you propmt to use SCSS our CSS solution is also OK?
> 
> I think it would nice to use SCSS - it is used way more common than classic CSS

From readme.md I understood that I can, but I don't have to use scss.

choeqq:
Sorry for misunderstanding, and thank you for feedback - we need to update readme!

Please provide solution using SCSS 😉

src/blocks/card.scss
```diff
@@ -0,0 +1,92 @@
...
+  &__image {
+    margin-top: 16px;
+    width: 160px;
+    height: 134px;
+
+  }
+
+  &__product-title {
+    margin-top: 40px;
+    color: var(--main-accent, #060b35);
```
asakevych, line 22: 1.  please use scss variables - `$main-accent`
2. move all variables to seperate file

src/blocks/card.scss
```diff
@@ -0,0 +1,92 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  box-sizing: border-box;
+  width: 200px;
+  height: 408px;
```
asakevych, line 7: it's better not to use height here, let's make content dedicate it 😉

asakevych (changes requested):
So far so good!! small changes left 😉

choeqq (changes requested):
Waiting for any updates @Sancz0pansa 👀

choeqq (approved):
All good! Approved ✅

PR #3271, opened 2023-07-05
add card component

Meg-Sowka:
- [DEMO LINK](https://Meg-Sowka.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://Meg-Sowka.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -5,9 +5,46 @@
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Product cards</title>
-    <link rel="stylesheet" href="./style.css">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+    <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <img
+        src="./images/imac.jpeg"
+        alt="APPLE A1419 iMac"
+        class="card__product-preview">
+      <div class="card__info">
+        <h2 class="card__product-name">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </h2>
+        <p class="card__product-code">
+          Product code: 195434
+        </p>
+        <div class="card__rates">
+          <div class="stars stars--4 card__stars">
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+          </div>
+          <div class="card__reviews">Reviews: 5</div>
+        </div>
+        <div class="card__price">
+          <span class="card__price-txt">Price:</span>
+          <span class="card__price-value">$2,199</span>
+        </div>
+      </div>
+      <div class="card__action-row">
+        <a
+          href="/productLink"
+          class="card__link"
+          data-qa="hover"
+        >
+          Buy
+        </a>
+      </div>
+    </div>
   </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -12,7 +12,7 @@
     <div class="card" data-qa="card">
       <img
         src="./images/imac.jpeg"
-        alt="APPLE A1419 iMac"
+        alt="APPLE A1419 iMac 27 Retina 5K Monoblock (MNED2UA/A)"
         class="card__product-preview">
       <div class="card__info">
         <h2 class="card__product-name">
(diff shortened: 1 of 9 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,117 @@
...
+    height: $gap;
+    margin-right: 4px;
+  }
+
+  @for $i from 1 through 5 {
+    &--#{$i} &__star:nth-child(-n+#{$i}) {
+      background-image: url(../images/star-active.svg);
+    }
+  }
+}
```
choeqq, line 117: really nice and according to [checklist](https://github.com/mate-academy/layout_product-cards/blob/master/checklist.md) point 11 - we want to have one file for every BEM block - please extract it to another file

src/styles/main.scss
```diff
@@ -0,0 +1,117 @@
...
+  background-color: $color;
+  border: 1px solid $borderCol;
+  border-radius: 5px;
+}
+
+@mixin font($weight, $size, $line: $size) {
+  font-size: $size;
+  line-height: $line;
+  font-weight: $weight;
+}
```
choeqq, line 19: maybe instead of using function you can set the font in some top level tag, like body

src/styles/main.scss
```diff
@@ -0,0 +1,117 @@
...
+$cardBorder: #f3f3f3;
+$cardBtnColor: #00acdc;
+$gap: 16px;
+$mainFont: "Roboto", sans-serif;
+
+@mixin base($color, $borderCol) {
+  background-color: $color;
+  border: 1px solid $borderCol;
+  border-radius: 5px;
+}
```
choeqq, line 13: I think you're using this function only once, so there is not point of declaring function - it's better to keep it as simple as possible

src/styles/main.scss
choeqq, line 1: please follow [checklist](https://github.com/mate-academy/layout_product-cards/blob/master/checklist.md) point 11 for better file structure

src/index.html
```diff
@@ -5,9 +5,46 @@
...
-    <link rel="stylesheet" href="./style.css">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+    <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <img
+        src="./images/imac.jpeg"
+        alt="APPLE A1419 iMac"
```
choeqq, line 15: nice, but I think you can provide even more descriptive alt - [checklist](https://github.com/mate-academy/layout_product-cards/blob/master/checklist.md) point 13

src/index.html
```diff
@@ -5,9 +5,46 @@
...
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+          </div>
+          <div class="card__reviews">Reviews: 5</div>
+        </div>
+        <div class="card__price">
+          <span class="card__price-txt">Price:</span>
+          <span class="card__price-value">$2,199</span>
```
choeqq, line 36:
```suggestion
          <span class="card__price-text">Price:</span>
          <span class="card__price-amount">$2,199</span>
```
like that it's more explicit 😄

src/index.html
```diff
@@ -5,9 +5,46 @@
...
+        </div>
+      </div>
+      <div class="card__action-row">
+        <a
+          href="/productLink"
```
choeqq, line 41: typicially we use `/product-link` notation for urls

src/styles/main.scss
```diff
@@ -0,0 +1,117 @@
...
+  line-height: $line;
+  font-weight: $weight;
+}
+
+*,
+*::after,
+*::before {
+  box-sizing: border-box;
+  margin: 0;
+}
```
choeqq, line 26: you can use that in index.scss file, after creating files for card and stars

choeqq (changes requested):
Just small adjustments needed 🚀

src/styles/main.scss
Meg-Sowka: due to many different font styles, even after adding it to a body I still use it around 4 times so I decided to leave it :)

src/styles/main.scss
Meg-Sowka: this one is actually used twice, once for card itself, second for button :)

src/index.html
Meg-Sowka: naming variables and classes are my absolute weakness 😅

src/styles/main.scss
Meg-Sowka: I will move css blocks to separate files and leave those lines in main.scss, under imports. ( I usually not using "index" name for scss) :)

asakevych (approved):
Great job!! Perfect solution!! 🚀 🔥 🔥 👏

PR #3273, opened 2023-07-05
Solution

Erchu420:
- [DEMO LINK](https://Erchu420.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://Erchu420.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -5,9 +5,61 @@
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Product cards</title>
-    <link rel="stylesheet" href="./style.css">
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="anonymous">
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
+      rel="stylesheet">
+    <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <img
+        src="images/imac.jpeg"
+        alt="Apple iMac 27&quot;"
+        class="card__img">
+
+      <p class="card__name">
+          APPLE A1419 iMac 27" Retina <br>
+          5K Monoblock (MNED2UA/A)
+      </p>
+
+      <p class="card__code">
+        Product code: 195434
+      </p>
+
+      <div class="card__votes">
+        <div class="card__stars stars stars--4">
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+        </div>
+
+        <p class="card__reviews">
+          Reviews: 5
+        </p>
+      </div>
+
+      <div class="card__price">
+        <p class="card__price-label">
+          Price:
+        </p>
+
+        <p class="card__price-value">
+          $2,199
+        </p>
+      </div>
+
+      <a
+        class="card__button-buy"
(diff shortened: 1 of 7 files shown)


src/index.html
```diff
@@ -5,9 +5,61 @@
...
+          Price:
+        </p>
+
+        <p class="card__price_value">
+          $2,199
+        </p>
+      </div>
+
+      <a
+        class="card__button_buy"
```
Oskinavara, line 58: This naming is wrong, classnames in html should be written in kebab-case, so it should be `card__button-buy` instead

Oskinavara: It should be changed in other classes as well, for example `card__price_label` is also wrong

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,95 @@
...
+
+* {
+  box-sizing: border-box;
+}
+
+.card {
+  box-sizing: border-box;
+  font-family: "Roboto", Helvetica, sans-serif;
+  width: 200px;
+  height: 408px;
```
Oskinavara, line 11: Setting fixed width is not the best idea, it would be better to set it to 100%, and sex max-width, so it can stay responsive and shrink if needed. Setting fixed height is really bad idea. What happens if the text wraps? The content will now overflow the container. Height should be determined by contents.

Oskinavara (changes requested):

src/index.html
Erchu420: updated in 'solution2'

src/styles/blocks/card.scss
Erchu420: updated in 'solution2'

src/index.html
```diff
@@ -5,9 +5,61 @@
...
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
+      rel="stylesheet">
+    <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <img
+        src="images/imac.jpeg"
+        alt="Apple iMac 27&quot;"
```
choeqq, line 21: please provide more descriptive alt - [checklist ](https://github.com/mate-academy/layout_product-cards/blob/master/checklist.md)point 13

src/index.html
```diff
@@ -5,9 +5,61 @@
...
+        class="card__img">
+
+      <p class="card__name">
+          APPLE A1419 iMac 27" Retina <br>
+          5K Monoblock (MNED2UA/A)
+      </p>
```
choeqq, line 27: I think I would use `<h2>` here

src/index.html
```diff
@@ -5,9 +5,61 @@
...
+
+      <p class="card__code">
+        Product code: 195434
+      </p>
+
+      <div class="card__votes">
```
choeqq, line 33:
```suggestion
      <div class="card__reviews">
```
I think reviews sounds better 😄

src/styles/blocks/stars.scss
choeqq, line 1: nicely done!

src/styles/utilies/_mixin.scss
choeqq, line 1: I don't think this function is necessary, it makes our code llittle harder to read without many benefits - I think I would stick to setting it all manually - for readibility

choeqq (changes requested):
Almost there, just little changes needed 🚀

src/index.html
Erchu420: updated in solution3

src/index.html
Erchu420: updated in solution3

src/index.html
Erchu420: updated in solution3

asakevych (approved):
Great job!!! 🚀 🔥 👏 👏

PR #3276, opened 2023-07-06
add card solution

vladyslav-svorin:
- [DEMO LINK](https://vladyslav-svorin.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://vladyslav-svorin.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,30 +11,30 @@
   <body>
     <main class="page">
       <div data-qa="card" class="page__card card">
-        <div class="card__image"></div>
+        <div class="card__image">
+          <img src="./images/imac.jpeg" alt="APPLE A1419 iMac 27 Retina 5K Monoblock (MNED2UA/A)">
+        </div>
 
-        <div class="card__info info-card">
-          <div class="info-card__main main-info">
-            <h2 class="main-info__title">APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</h2>
-            <div class="main-info__poduct-code">Product code: 195434</div>
+        <div class="card__main">
+          <h2 class="card__title">APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</h2>
+          <div class="card__poduct-code">Product code: 195434</div>
+        </div>
+
+        <div class="card__review">
+          <div class="card__raiting stars stars--4">
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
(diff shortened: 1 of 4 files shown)


src/index.html
```diff
@@ -5,9 +5,46 @@
...
     <title>Product cards</title>
-    <link rel="stylesheet" href="./style.css">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+    <link rel="stylesheet" href="./styles/index.scss">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <main class="page">
+      <div data-qa="card" class="page__card card">
+        <div class="card__image"></div>
```
TarasHoliuk, line 14: it's a content image, so use `img` tag to add it instead of background-image

src/index.html
```diff
@@ -5,9 +5,46 @@
...
+
+        <div class="card__info info-card">
```
TarasHoliuk, line 16:
```suggestion
        <div class="card__info">
```
it's just an element even if contains nested elements
P.S. This wrapper seems redundant to me - just add a bigger margin to the Price block

src/index.html
```diff
@@ -5,9 +5,46 @@
...
+          <div class="info-card__main main-info">
```
TarasHoliuk, line 17:
```suggestion
          <div class="card__main-info">
```
it's also just an element of `card` block even if it's inside another card's element, i.e. that's allowed:
```
<block>
  <block__element1>
    <block__element2 />
  </block__element1>
</block>
```

src/index.html
TarasHoliuk, line 14: FYI you may use self-closing tag
```suggestion
        <div class="card__image" />
```

src/index.html
TarasHoliuk: fix everywhere - you should have only 2 blocks here - card and stars
P.S. This wrapper seems redundant to me - just add a bigger margin to `Product code`

src/styles/index.scss
```diff
@@ -1,3 +1,149 @@
...
+
+.page {
+
+  &__card {
+    display: flex;
+    max-width: 200px;
+    flex-direction: column;
+    background-color: #fff;
+    border: 1px solid #f3f3f3;
+    padding: 32px 16px 16px;
```
TarasHoliuk, line 24: all these styles might be applied to `card` as to BEM-block - BEM-mix with `page` block is redundant here (as page block at all)

TarasHoliuk (changes requested):

sTorba24 (approved):
Good work!

PR #3287, opened 2023-07-07
Product layout card finish

WaveOfDandelions:
- [DEMO LINK](https://WaveOfDandelions.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://WaveOfDandelions.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -4,10 +4,56 @@
     <meta charset="UTF-8">
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+    <link rel="stylesheet" href="./styles/index.scss">
     <title>Product cards</title>
-    <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <main class="page">
+      <div class="card" data-qa="card">
+        <div class="card__item item">
+
+          <div class="item__image-container">
+            <img class="item__image" src="./images/imac.jpeg"
+              alt="APPLE A1419 iMac 27 Retina 5K Monoblock">
+          </div>
+
+          <h2 class="item__name">
+            APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+          </h2>
+          <p class="item__code">Product code: 195434</p>
+
+          <div class="item__rate-container rate">
+            <div class="rate__stars-container stars rate__stars-container_stars--1">
+              <ul class="list stars__list">
+                <li class="list__item"></li>
+                <li class="list__item"></li>
+                <li class="list__item"></li>
+                <li class="list__item"></li>
+                <li class="list__item"></li>
+              </ul>
+            </div>
+            <div class="rate__review review">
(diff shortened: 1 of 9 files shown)


src/index.html
```diff
@@ -4,10 +4,56 @@
...
+            </div>
+          </div>
+
+          <div class="item__price-container price">
+            <p class="price__text">Price:</p>
+            <p class="price__value">$2,199</p>
+          </div>
+
+          <button class="item__button" type="submit">
+            <a class="item__link" href="/productLink"
```
Oskinavara, line 49: Formatting is really weird here. Please use the autoformatter

src/styles/blocks/_card.scss
```diff
@@ -0,0 +1,78 @@
+.card {
+  width: 198px;
+  @include border(1px, solid, $border-color, 5px);
+  .card__item {
```
Oskinavara, line 4: Such nesting is unnecessary and increases the specificity of the selector which is generally bad. Use SCSS nesting instead.

src/styles/blocks/_card.scss
```diff
@@ -0,0 +1,78 @@
...
+      margin-bottom: 16px;
+      .price__text {
+        @include font(400, 12px, 18px, $font-color--secondary, Roboto);
+      }
+      .price__value {
+        @include font(700, 16px, 18px, $font-color, Roboto);
+      }
+    }
+    .item__button {
+      height: 40px;
```
Oskinavara, line 55: Setting fixed widths and heights is generally a bad idea. Width of the element can be determined by it's parent and set to 100%. Height could be determined by line-height + paddings.

src/styles/blocks/_card.scss
```diff
@@ -0,0 +1,78 @@
+.card {
+  width: 198px;
```
Oskinavara, line 2: It would be better to have max-width: 198px and width: 100%; so it can be narrower if needed

Oskinavara (changes requested):

src/index.html
WaveOfDandelions: > Formatting is really weird here. Please use the autoformatter

I use htmlint and editor config with 2 spaces, can you explaine me how it should be?

src/styles/blocks/_card.scss
WaveOfDandelions: > Such nesting is unnecessary and increases the specificity of the selector which is generally bad. Use SCSS nesting instead.

If I nest like this .card__item inside .card it is not scss nesting??

choeqq (changes requested):
Waiting for any updates @WaveOfDandelions 👀

src/index.html
WaveOfDandelions: > Formatting is really weird here. Please use the autoformatter

Fixed

src/styles/blocks/_card.scss
WaveOfDandelions: > Such nesting is unnecessary and increases the specificity of the selector which is generally bad. Use SCSS nesting instead.

Done

src/styles/blocks/_card.scss
WaveOfDandelions: > Setting fixed widths and heights is generally a bad idea. Width of the element can be determined by it's parent and set to 100%. Height could be determined by line-height + paddings.

I was change properties for better flexibility

src/styles/blocks/_card.scss
WaveOfDandelions: > It would be better to have max-width: 198px and width: 100%; so it can be narrower if needed

Done

choeqq (approved):
Looks good ✅

PR #3313, opened 2023-07-11
add task solution

arturwisnia:


src/index.html
```diff
@@ -5,9 +5,56 @@
...
+
+      <h2 class="card__title">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </h2>
+
+      <p class="card__code">
+        Product code: 195434
+      </p>
+
+      <div class="card__rev">
```
asakevych, line 28: no need to use shorter classes, it would be easier to read when you use whole words -> 
change to card__reviews with card__stars and card__reviews-label inside 
or card__rating with card__stars and card__reviews inside

src/index.html
```diff
@@ -5,9 +5,56 @@
...
+      </div>
+
+      <div class="card__price">
+        <h2 class="card__price-start">
+          Price:
+        </h2>
+
+        <h2 class="card__price-end">
+          $2,199
+        </h2>
```
asakevych, line 49: 1. no need to use h2 here, it's not a title, just some tekst with information, you can use span or p instead
2. start and end are not hte best names, it's better to use more valuable names, it can be `card__price-label` and `card__price-value` or `card__price-amount`

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,93 @@
...
+    align-items: center;
+    margin: 0 0 24px;
+  }
+
+  &__rev-text {
+    font-weight: 400;
+    font-size: 10px;
+    line-height: 14px;
+    color: $main;
+    margin: 0;
```
asakevych, line 48: 1. please set margin 0 for all p elements instead of give this property to every class of p element 
2. you can create file normalize scss in which you write all such styles like *, :root, body, p etc. -> it's a good practice to reset such default styles of browsers in separate file and set some general settings there

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,93 @@
...
+    justify-content: space-between;
+    align-items: center;
+    margin-bottom: 16px;
+
+    &-start {
+      font-size: 12px;
+      font-weight: 400;
+      line-height: 18px;
+      color: $secondary;
+      margin: 0;
```
asakevych, line 62: 1. same with resetting margins for all p elements
2. use another name for this class

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    &-end {
+      font-size: 16px;
+      font-weight: 700;
+      line-height: 18px;
+      color: $main;
+      margin: 0;
```
asakevych, line 70: 1. same with resetting margins for all p elements
2. use another name for this class

src/index.html
```diff
@@ -5,9 +5,56 @@
...
+      </div>
+
+      <a
+        class="card__button"
+        href="#"
+        data-qa="hover">
```
asakevych, line 55: closing tag should from new line

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,93 @@
...
+    height: 40px;
+    display: block;
+    background-color: $blue;
+    color: white;
+    text-decoration: none;
+    border-radius: 5px;
+    padding: 12px 70px 12px 69px;
+    font-size: 14px;
+    line-height: 16px;
+    font-weight: 700;
```
asakevych, line 85: there is no need to set other display, paddings and height -> just use `line-height: 40px` and `align-text: center` 😉

src/styles/index.scss
```diff
@@ -1,3 +1,11 @@
 body {
   margin: 0;
```
asakevych, line 2: you can add font-family here also

asakevych (changes requested):
Good job!! just some small changes are needed 😉

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,88 @@
...
+  flex-direction: column;
+  border-radius: 5px;
+  border: 1px solid $white;
+  padding: 32px 16px 16px;
+
+  &__img {
+    width: 160px;
+    height: 134px;
+    margin-bottom: 40px;
+    align-items: center;
```
asakevych, line 13: it ahs no sense here, this property should be on parent element -> `. card`

asakevych (changes requested):
please check all comments once more 😉

src/index.html
mvjl000, line 1: The way you are using the section heading elements is not ideal. Take a look at [the docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements). Those tags are supposed to indicate new sections. Instead of using h2 in title and price it would be better to simply use a `p` elements

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,84 @@
+.card {
+  width: 200px;
+  display: flex;
+  flex-direction: column;
+  border-radius: 5px;
+  border: 1px solid $white;
+  padding: 32px 16px 16px;
+  margin-bottom: 40px;
```
mvjl000, line 8: This card doesn't need the margin-bottom. It shouldn't be in block styles.

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,21 @@
+.stars {
+  display: flex;
+
+  &__star {
+    width: 16px;
+    height: 16px;
```
mvjl000, line 6: Same value being used. Good use case for a variable ;)

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,84 @@
...
+      line-height: 18px;
+      color: $secondary;
+    }
+
+    &-second {
+      font-size: 16px;
+      font-weight: 700;
+      line-height: 18px;
+      color: $main;
+    }
```
mvjl000, line 63: Variable for line-height

mvjl000 (changes requested):

PR #3318, opened 2023-07-13
create a product card

FrontKid:


src/styles/1-helpers/_mixins.scss
```diff
@@ -0,0 +1,5 @@
+@mixin flex-centering() {
+  display: flex;
+  justify-content: space-between;
```
roman-mirzoian, line 3: Why is your mixin called `flex-centering` if you do `justify-content: space-between`? 
It would be better then to call it some type of `flex-alignment`.

src/index.html
```diff
@@ -5,9 +5,62 @@
...
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+    <link rel="stylesheet" href="./styles/index.scss">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <img
+      class="card__img"
+      src="./images/imac.jpeg"
+      alt="APPLE A1419 iMac 27&quot; Retina<br>5K Monoblock (MNED2UA/A)"
```
roman-mirzoian, line 22: `alt` attribute of `img` tag should not include any HTML tags like `<br>`. 
It should be a plain text description of the image, used by screen readers and when the image can't be loaded.

src/index.html
```diff
@@ -5,9 +5,62 @@
...
     <title>Product cards</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="anonymous"
+      >
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+    <link rel="stylesheet" href="./styles/index.scss">
```
roman-mirzoian, line 15: It is a bad practice to connect the SCSS file directly in HTML file.
Please compile your SCSS into a CSS file, and link the CSS file in your HTML instead.

src/index.html
```diff
@@ -5,9 +5,62 @@
...
+      <img
+      class="card__img"
+      src="./images/imac.jpeg"
+      alt="APPLE A1419 iMac 27&quot; Retina<br>5K Monoblock (MNED2UA/A)"
+      width="160"
+      height="134"
+      >
+
+      <h3 class="card__title">
+        APPLE A1419 iMac 27&quot; Retina<br>5K Monoblock (MNED2UA/A)
```
roman-mirzoian, line 28: You have used a `<br>` tag in `h3` heading. It's better to avoid using `<br>` tags in headings or paragraph text. Consider using CSS to control the design or layout such as `width` or `height` to enforce line breaks where necessary.

src/styles/1-helpers/_variables.scss
```diff
@@ -0,0 +1,9 @@
+// colors
+$white: #fff;
+$blue: #00acdc;
+$black: #060b35;
+$grey: #616070;
```
roman-mirzoian, line 5: It is better to give names so that their purpose is clear, and not just to repeat the name of the color, for example.
Because the color and other values may change during development:
```
$cardBackgroundColor: #fff;
$buttonColor: #00acdc;
etc.
```

src/styles/1-helpers/_variables.scss
```diff
@@ -0,0 +1,9 @@
...
+
+// general
+$fz: 12px;
+$lh: 18px;
```
roman-mirzoian, line 9: The same applies to these variables. You should not give them unreadable names, because now without context it is not clear what is stored in these variables.

src/styles/1-helpers/_common.scss
```diff
@@ -0,0 +1,8 @@
+html {
+  font-size: $fz;
+  font-family: "Roboto", "Helvetica", sans-serif;
```
roman-mirzoian, line 3: The `font-family` value should also be stored in a variable.

roman-mirzoian (changes requested):
Good work, but it needs to be improved a little 👍

src/index.html
FrontKid: it was written in the task description:
https://github.com/FrontKid/layout_product-cards/tree/develop#the-task

if i compiled in css there will be no spaces between the lines, I don't know how to set it up in the plugin

src/index.html
```diff
@@ -5,9 +5,62 @@
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Product cards</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="anonymous"
+      >
```
BudnikOleksii, line 13:
```suggestion
    <link
      rel="preconnect"
      href="https://fonts.gstatic.com"
      crossorigin="anonymous"
    >
```

src/index.html
```diff
@@ -5,9 +5,62 @@
...
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <img
+      class="card__img"
+      src="./images/imac.jpeg"
+      alt="APPLE A1419 iMac 27&quot; Retina 5K Monoblock (MNED2UA/A)"
+      width="160"
+      height="134"
+      >
```
BudnikOleksii, line 25:
```suggestion
      <img
        class="card__img"
        src="./images/imac.jpeg"
        alt="APPLE A1419 iMac 27&quot; Retina 5K Monoblock (MNED2UA/A)"
      >
```
Don't use inline styles, specify properties in the css/scss files with class.
[CODE STYLE] - If the HTML-element has long attribute values or number of attributes is more than 2 - start each one, including the first, on the new line with 2-space indentation related to tag. Tag’s closing bracket should be on the same level as opening one.

src/styles/2-components/_card.scss
```diff
@@ -0,0 +1,53 @@
...
+
+  &__img {
+    align-self: center;
+    margin-bottom: 40px;
+  }
+
+  &__title {
+    font-weight: 500;
+    margin-bottom: 4px;
+    max-width: 166px;
```
BudnikOleksii, line 19: You don't need max-width in every element as your "card" have a fixed width 200px and paddings: 32px 16px 16px;

src/styles/index.scss
```diff
@@ -1,3 +1,8 @@
...
-  margin: 0;
-}
+@import "./1-helpers/variables";
+@import "./1-helpers/mixins";
+@import "./1-helpers/reset";
+@import "./1-helpers/common";
+
+@import "./2-components/card";
+@import "./2-components/stars";
+@import "./3-UI/card-button";
```
BudnikOleksii, line 8: It's not so important in this case, but why do you name folders with such a pattern as "1-helpers"?

BudnikOleksii (changes requested):
Great work👍🏻
Just fix your project slightly. Let's get rid of unnecessary max-width properties.
8. [STYLES] - Don't hardcode exact height size. Add necessary paddings according to mockup and let content dictate the container size.

src/styles/index.scss
FrontKid: I thought about how to structure my project into folders and looked at the real world example there I saw a similar structure and I liked it

src/index.html
```diff
@@ -5,9 +5,62 @@
...
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <img
+        class="card__img"
+        src="./images/imac.jpeg"
+        alt="APPLE A1419 iMac 27&quot; Retina 5K Monoblock (MNED2UA/A)"
+        width="160"
+        height="134"
```
BudnikOleksii, line 24: Don't use inline styles

BudnikOleksii (changes requested):
Nice, but remove inline styles.

src/index.html
FrontKid: fixed

roman-mirzoian (changes requested):

image

src/styles/1-helpers/_variables.scss
```diff
@@ -0,0 +1,14 @@
...
+$backgroundButtonColor: #00acdc;
+$cardTextColor: #060b35;
+$subTitleColor: #616070;
+
+// general
+$mainFontSize: 12px;
+$mainLineHeight: 18px;
+
+// font-fa
+
```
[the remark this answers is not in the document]
FrontKid: done

tiserett (approved):
LGTM 🔥🚀

Try to use more semantic tags

PR #3322, opened 2023-07-13
card was created

maksym-mishchanchuk:


src/index.html
```diff
@@ -5,9 +5,45 @@
...
+      data-qa="card"
+    >
+      <img
+        class="card__img"
+        src="../src/images/imac.jpeg"
+        alt="APPLE A1419 iMac 27 Retina 5K Monoblock"
+      >
+      <h2 class="card__name">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </h2>
```
alexander-ignatow, line 25: I would not use `hx` tags in such a task.

Let's make it clear - in the context of your task it is ok, as you have a single product card per your entire page. But imagine a real-world store with multiple product cards on the page - that will lead to dozens `h2`s and even worse dozens of _the same_ `h3`'s .  The latter would almost definitely hurt SEO and page readability for screen readers.

Unfortunately, I am not a SEO specialist to give you a deep explanation, but most probably you should avoid using `hx` tags in such blocks

Probably this article will give you more idea:
https://semji.com/blog/why-use-h1-h2-h3-tags-for-your-seo

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,94 @@
+.card {
+  width: 200px;
+  box-sizing: border-box;
+  padding: 32px 16px 16px;
+  border-radius: 5px;
+  border: 1px solid #f3f3f3;
```
alexander-ignatow, line 6: Hardcoded color - please create variable for it too

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,20 @@
...
+  }
+}
+
+.stars--1 > :first-child,
+.stars--2 > :nth-child(-n + 2),
+.stars--3 > :nth-child(-n + 3),
+.stars--4 > :nth-child(-n + 4),
+.stars--5 > :nth-child(-n + 5) {
+  background-image: url(../images/star-active.svg);
+}
```
alexander-ignatow, line 20: With SCSS you can improve this part using @for (hint - first you need to replace `first-child` with `nth-child`)
https://sass-lang.com/documentation/at-rules/control/for/

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,94 @@
+.card {
+  width: 200px;
+  box-sizing: border-box;
```
alexander-ignatow, line 3: Thats correct - you need to set this to make sure all works correctly
Still, a better idea is to reset `box-sizing` for all elements unless you have a reason not to do it (having in mind you set same border-box value for several elements)

https://css-tricks.com/box-sizing/#aa-good-better-and-probably-best-box-sizing-reset-methods

alexander-ignatow (changes requested):
Good start - no lets improve your solution together!

- per task you need to `reset browser's default margins`.
- Fix some minor issues below.

maksym-mishchanchuk:
> Good start - no lets improve your solution together!
> 
> * per task you need to `reset browser's default margins`.
> * Fix some minor issues below.

i have normalize file for this

dffUqp (approved):
LGTM🔥

PR #3337, opened 2023-07-18
add demo, first card

o-cherpak:
- [DEMO LINK](https://o-cherpak.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://o-cherpak.github.io/layout_product-cards/report/html_report/)

Why do I have incorrect indents when the paddings are specified correctly?

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,8 +6,40 @@
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Product cards</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect"
+      href="https://fonts.gstatic.com"  >
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div
+    class="card"
+    data-qa="card">
+
+      <img
+      src="images/imac.jpeg"
+      alt="card__product"
+      class="card__product">
+
+      <h1 class="card__title">APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</h1>
+
+      <p class="card__product-code">Product code: 195434</p>
+
+      <div class="stars stars--4">
+        <span class="stars__star"></span>
+        <span class="stars__star"></span>
+        <span class="stars__star"></span>
+        <span class="stars__star"></span>
+        <span class="stars__star"></span>
+
+        <p class="stars__review">Reviews: 5</p>
+      </div>
+
+      <p class="card__price">Price:</p>
+
+      <h2 class="card__title-price">$2,199</h2>
+
+      <a href="#buy" class="button">Buy</a>
+    </div>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -6,8 +6,40 @@
...
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div
+    class="card"
+    data-qa="card">
+
+      <img
+      src="images/imac.jpeg"
+      alt="card__product"
```
Esceype, line 21: Checklist point number 13

src/index.html
```diff
@@ -6,8 +6,40 @@
...
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect"
+      href="https://fonts.gstatic.com"  >
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div
+    class="card"
+    data-qa="card">
```
Esceype, line 17: Checklist point number 14

Esceype (changes requested):
Nice work but need to make some changes, read the following comments and make the appropriate changes

o-cherpak:
How can I set the size of the elements, in particular the card without specifying the width and height?

o-cherpak:
@volodymyr-soltys97

src/style.css
```diff
@@ -1,3 +1,84 @@
...
+a,
+h1,
+h2,
+img,
+p {
   margin: 0;
+  padding: 0;
+}
+
+.card {
```
xiimiik, line 23: It is necessary to set the width for the card, but the height will be expanded by the content, because if we do not limit the width, it will be full screen, as it is now

src/style.css
```diff
@@ -1,3 +1,84 @@
...
+  box-sizing: border-box;
+  border-radius: 5px;
+  border: 1px solid #f3f3f3;
+  background: #fff;
+}
+
+.card__imac {
```
xiimiik, line 30: you can also set constant sizes for images

xiimiik (changes requested):

Andriy-Shymkiv (changes requested):
- this behaviour is not correct. you need just to give `margin-bottom`:
image
- same here:
image
- you gave horizontal indent almost for every card item. it's not correct, and such code is really hard to support, so please give `padding` for `card` instead of it and remove horizontal margins for card items:
image

src/style.css
```diff
@@ -1,3 +1,108 @@
...
+a,
+h1,
+h2,
+img,
+p {
   margin: 0;
+  padding: 0;
+}
+
+.card {
```
polosanya, line 23: Create a separate file for card block
image

src/style.css
```diff
@@ -1,3 +1,108 @@
...
+  color: var(--blue-accent);
+  text-align: center;
+  font-size: 14px;
+  font-weight: 700;
+  line-height: 38px;
+  text-transform: uppercase;
+  text-decoration: none;
+  border-radius: 5px;
+  background: #fff;
+  border: 1px solid var(--blue-accent);
```
polosanya, line 107: There is no need to duplicate all properties. Set only the properties that change on hover

polosanya (changes requested):
Almost done! 
Feel free to ask questions in fe_chat if you have any difficulties)

o-cherpak:
@polosanya @Andriy-Shymkiv 
Please check because I will lose my streak

xiimiik (approved):

PR #3357, opened 2023-07-19
create card

KatOlista:
- [DEMO LINK](https://KatOlista.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://KatOlista.github.io/layout_product-cards/report/html_report/)


src/styles/index.scss
```diff
@@ -1,3 +1,135 @@
...
+}
+
+.card {
+  padding: 31px 0 15px;
+  $blue-accent: #00acdc;
+  $white-accent: #fff;
+  display: flex;
+  align-items: center;
+  flex-direction: column;
+  font-family: "Roboto", sans-serif;
```
dffUqp, line 13: better to define global styles in `body`.

move font-family to the `body`

dffUqp (changes requested):
LGTM:fire: but needs a few minor fixes:smiley:

src/index.html
```diff
@@ -5,9 +5,39 @@
...
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div data-qa="card" class="card">
+      <img
+        src="./images/imac.jpeg"
+        alt="mac-picture"
+        class="card__image"
+      >
```
BudnikOleksii, line 17:
```suggestion
      <img
        src="./images/imac.jpeg"
        alt="APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)"
        class="card__image"
      >
```
Checklist:
13. [SEO] - alt atribute should describe the image if the image contains information (better description you have - better for you :))

src/styles/index.scss
```diff
@@ -1,3 +1,132 @@
+$hoverBlueColor: #00acdc;
+$hoverWhiteColor: #fff;
+
 body {
   margin: 0;
+  font-family: "Roboto", sans-serif;
```
BudnikOleksii, line 6: Checklist:
image

src/styles/index.scss
```diff
@@ -1,3 +1,132 @@
...
+
+.card {
+  padding: 32px 0 16px;
+  display: flex;
+  align-items: center;
+  flex-direction: column;
+  font-size: 10px;
+  line-height: 14px;
+  font-weight: 400;
+  width: 198px;
```
BudnikOleksii, line 17: You can use box-sizing: border-box; property and don't calculate manually width with borders

BudnikOleksii (changes requested):
Looks great🚀
Just a few things to improve your solution.
Also, let's create separate files for the BEM blocks, checklist:
11. [FILE STRUCTURE] - You should create separate file for each BEM block

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,2 @@
+$hoverBlueColor: #00acdc;
+$hoverWhiteColor: #fff;
```
roman-mirzoian, line 2: You need to fix the variable names, they should show their purpose, not their content. if I decided to change the color, I would have to change it everywhere, not just in the variable definition.

roman-mirzoian (changes requested):
I also see that the tests are not passing now, this must be fixed:
image

KatOlista:
>  I have spent a lot of time trying to fix my code and pass tests. But I can't solve that problem. All tags are in the correct places.
I used the correct colors. But the test always makes a screenshot with not my text color tone. I have checked the colors in another student's solves - they use the same. You can see a difference in the attached screenshot

image

src/styles/components/_card.scss
```diff
@@ -0,0 +1,88 @@
...
+  border-radius: 5px;
+  border: 1px solid #f3f3f3;
+  background: #fff;
+  color: $primeColor;
+  line-height: 14px;
+
+  @include fontParams(10px, 400);
+
+  &__image {
+    margin: 0 0 40px;
```
BudnikOleksii, line 17: If you need only one margin property it's better to use what you need instead of shorthand property.

src/styles/components/_card.scss
```diff
@@ -0,0 +1,88 @@
...
+    color: $secondaryColor;
+    font-size: 12px;
+  }
+
+  &__info-amount {
+    @include fontParams(16px, 700);
+  }
+
+  &__link {
+    height: 40px;
```
BudnikOleksii, line 68: Checklist:
8. [STYLES] - Don't hardcode exact height size. Add necessary paddings according to mockup and let content dictate the container size.

src/styles/components/_stars.scss
```diff
@@ -0,0 +1,29 @@
+.stars {
+  display: flex;
+  margin: 0;
+  padding: 0;
```
BudnikOleksii, line 4: If you want to reset default browser styles it's better to do it for tags and in reset.scss/_normillize.scss file.

src/styles/index.scss
```diff
@@ -1,3 +1,6 @@
-body {
-  margin: 0;
-}
+@import url("https://fonts.googleapis.com/css2?family=Manrope:wght@400;700&family=Roboto:wght@400;500;700&display=swap");
```
BudnikOleksii, line 1: I don't see in your code where you use "Manrope" font family, if you don't use it remove it.

BudnikOleksii (changes requested):
I've checked the demo and compared it with the design in Figma, colors and fonts are the same, looks good, let's skip this problem with tests.
Let's fix the hardcoded height size for link and remove unused font style import.

src/styles/components/_stars.scss
BudnikOleksii: Pay attention to comments

BudnikOleksii (approved):
👍

PR #3361, opened 2023-07-19
add task solution

OlhaMomot:
- [DEMO LINK](https://OlhaMomot.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://OlhaMomot.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Product cards
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_product-cards/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_product-cards/report/html_report/)
+- [DEMO LINK](https://OlhaMomot.github.io/layout_product-cards/)
+- [TEST REPORT LINK](https://OlhaMomot.github.io/layout_product-cards/report/html_report/)
 
 > Follow [this instruction](https://mate-academy.github.io/layout_task-guideline)
 ___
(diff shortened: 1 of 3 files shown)


src/styles/utils/_normalize.scss
```diff
@@ -0,0 +1,5 @@
+html {
+  font-family: "Roboto", sans-serif;
```
BudnikOleksii, line 2: Checklist:
image

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,4 @@
+$main-accent: #060b35;
+$blue-accent: #00acdc;
+$secondary: #616070;
```
BudnikOleksii, line 3: secondary what? Also, let's rename variables, they should show their purpose, not their content.

src/styles/card.scss
```diff
@@ -0,0 +1,88 @@
...
+    color: $secondary-font-color;
+    font-size: 12px;
+    font-weight: 400;
+  }
+
+  &__price-amount {
+    color: $main-font-color;
+    font-size: 16px;
+    font-weight: 700;
+  }
```
tiserett, line 66: nesting

src/styles/stars.scss
```diff
@@ -0,0 +1,19 @@
+.stars {
+  display: flex;
+  height: 16px;
```
tiserett, line 3: you don't need it here, your stars will define height of the block

tiserett (changes requested):
good job, but I found few minor problems ;)

src/styles/card.scss
OlhaMomot: I'm not sure how to do that... "price-title" and "price-amount", for example, are not elements of some "price" block. And we were told not to create an element of element. So I need to rename all that and create a bunch of new blocks in order to do nesting, don't I? If I understand the rule correctly.

BudnikOleksii (approved):
Nicely done🔥

PR #3376, opened 2023-07-20
add solution

Yevheniia-Sid:


src/styles/index.scss
```diff
@@ -1,3 +1,131 @@
...
+.card {
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  display: flex;
+  flex-direction: column;
+  width: 200px;
+  min-width: 200px;
+  padding: 32px 16px 16px;
+  box-sizing: border-box;
+  font-family: Roboto, Helvetica, sans-serif;
```
BudnikOleksii, line 24: You have a fixed width, you don't need min-width in this case.
The main font-family should apply to the html or body tag.

BudnikOleksii (changes requested):
Looks great🚀
Just a few fixes, one comment left and let's separate your styles to different files.
Checklist:
11. [FILE STRUCTURE] - You should create separate file for each BEM block

Also, add links to the demo and the test report to PR, it would be more convenient for reviewers.

src/styles/card.scss
```diff
@@ -0,0 +1,95 @@
+$first-color: #060b35;
+$second-color: #616070;
+$button-color: #00acdc;
+
+.card {
+  border: 1px solid #f3f3f3;
```
tiserett, line 6: separate file for this bem block

src/styles/card.scss
```diff
@@ -0,0 +1,95 @@
+$first-color: #060b35;
+$second-color: #616070;
```
tiserett, line 2: primary / secondary instead of first / second

src/styles/card.scss
```diff
@@ -0,0 +1,95 @@
...
+$button-color: #00acdc;
+
+.card {
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  display: flex;
+  flex-direction: column;
+  width: 200px;
```
tiserett, line 10: you don't have to set width here, let the content define total width

src/styles/card.scss
```diff
@@ -0,0 +1,95 @@
...
+  padding: 32px 16px 16px;
+  box-sizing: border-box;
```
tiserett, line 12: why don't you specify border-box for '*' selector?

src/styles/card.scss
```diff
@@ -0,0 +1,95 @@
...
+    margin-top: 4px;
+    color: $second-color;
+  }
+
+  &__review {
+    display: flex;
+    width: 100%;
+    justify-content: space-between;
+    align-items: center;
+    margin-top: 16px;
```
tiserett, line 44: create variable for this value

src/styles/card.scss
tiserett, line 2: separate file for variables

src/styles/card.scss
```diff
@@ -0,0 +1,95 @@
...
+    line-height: 18px;
+    color: $first-color;
+  }
+
+  &__button {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    margin-top: 16px;
+    height: 40px;
```
tiserett, line 78: don't set height, use paddings instead

src/styles/card.scss
Yevheniia-Sid: > separate file for this bem block

I didn't quite understand what to do. The card block is already in a separate card.scss file.

src/styles/card.scss
Yevheniia-Sid: > you don't have to set width here, let the content define total width

Could you help me? If I disable this property, everything looks broken.

src/styles/card.scss
Yevheniia-Sid: > don't set height, use paddings instead

ok, but what a reason?

BudnikOleksii (approved):
Nicely done🔥

PR #3379, opened 2023-07-20
Develop

bogemna:
- [DEMO LINK](https://bogemna.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://bogemna.github.io/layout_product-cards/report/html_report/)


src/index.html
```diff
@@ -5,9 +5,37 @@
...
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com"
+    crossorigin="anonymous">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <div class="card__img">
+        <img src="./images/imac.2690641a.jpeg" alt="apple">
```
BudnikOleksii, line 17: Checklist:
13. [SEO] - alt atribute should describe the image if the image contains information (better description you have - better for you :))
image

src/index.html
```diff
@@ -5,9 +5,37 @@
...
+      </div>
+      <div class="card__title">
+        <p>
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </p>
+      </div>
```
BudnikOleksii, line 23:
```suggestion
      <h2 class="card__title">
        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
      </h2>
```
Do you need additional element inside without any styles?
Also, looks like it's important info, it's better to use h2 or h3 in this case.

src/index.html
```diff
@@ -5,9 +5,37 @@
...
+      <div class="card__description">
+        <p>Product code: 195434</p>
+      </div>
+      <div class="card__review review">
+        <img src="./images/Stars.png" alt="start">
```
BudnikOleksii, line 28: You should use the "starts" block from the previous task, it shouldn't be just an image.
Checklist:
4. [TASK] - Make sure to reuse part of [Stars task](https://github.com/mate-academy/layout_stars) here, and keep it operational in case you change block modifier.

src/index.html
```diff
@@ -5,9 +5,37 @@
...
+        <p class="review__5">Reviews: 5</p>
+      </div>
+      <div class="card__price price">
+        <p class="price__price">Price:</p>
+        <h3 class="price__dollar">$2,199</h3>
+      </div>
+      <a
+      href="#"
+      class="card__btn"
+      data-qa="hover">Buy</a>
```
BudnikOleksii, line 38:
```suggestion
      <a
        href="#"
        class="card__btn"
        data-qa="hover"
      >
        Buy
      </a>
```
Checklist:
14. [CODE STYLE] - If the HTML-element has long attribute values or number of attributes is more than 2 - start each one, including the first, on the new line with 2-space indentation related to tag. Tag’s closing bracket should be on the same level as opening one.

src/styles/components/card.scss
```diff
@@ -0,0 +1,101 @@
...
+  border: 1px solid #f3f3f3;
+  background-color: $white-color;
+
+  &__img {
+    width: 160px;
+    height: 134px;
+    margin-left: 19px;
+    margin-right: 19px;
+    margin-top: 32px;
+    margin-bottom: 40px;
```
BudnikOleksii, line 16: Use shorthand margin/padding property when you need to specify a few of them.

src/styles/components/card.scss
```diff
@@ -0,0 +1,101 @@
...
+    height: 100%;
+    img {
+      width: 160px;
+      height: 134px;
+      display: block;
+      width: 100%;
+      object-fit: cover;
+    }
```
BudnikOleksii, line 24: Don't use tags for styling, use classes instead.
Tags are used only to reset or normalize styles.

src/styles/components/card.scss
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  &__title {
+    color: $text-color;
+    font-family: 'Roboto', sans-serif;
```
BudnikOleksii, line 29: The main font-family should be applied to the body or html element.
Also, checklist:
image

src/styles/components/card.scss
```diff
@@ -0,0 +1,101 @@
...
+      color: $button-color;
+      border: 1px solid $button-color;
+
+    }
+  }
+
+
+
+
+}
```
BudnikOleksii, line 101: Remove empty lines.

src/styles/components/main.scss
```diff
@@ -0,0 +1,5 @@
+*{
+  box-sizing: border-box;
+  margin: 0;
+  font-family: 'Roboto', sans-serif;
+}
```
BudnikOleksii, line 5: Don't use * selector, font-family specify into html or body, reset margins only for elements that really need it.

src/styles/components/details.scss
```diff
@@ -0,0 +1,31 @@
...
+  align-self: center;
+  color: $text-color;
+  text-align: right;
+  font-family: Roboto;
+  font-size: 10px;
+  font-style: normal;
+  font-weight: 400;
+  line-height: 14px;
+  padding-top: 1px;
+}
```
BudnikOleksii, line 31: The file has name "details.scss" but no details block was found 🤷🏻‍♂️

For this task you need only "stars" and "card" BEM blocks. Checklist:
6. [BEM] - Don't make BEM structure too complicated. You don't need to create separate blocks for anything other than card itself and stars.

BudnikOleksii (changes requested):
Not bad start, but a lot have to be done)
Check comments and pay attention to the [checklist](https://github.com/mate-academy/layout_product-cards/blob/master/checklist.md).

src/styles/components/price.scss
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  &__number {
+    margin-bottom: 16px;
+    font-size: 16px;
+    @include text-bold;
+  }
+}
+
+.reviews {
```
roman-mirzoian, line 15: The `reviews` does not belong to the `price` block, so you should not keep them in one file.

src/styles/components/main.scss
```diff
@@ -0,0 +1,20 @@
+* {
+  box-sizing: border-box;
+}
+
+html {
```
roman-mirzoian, line 5: It is worth avoiding working with tags directly and setting styles more than classes.

roman-mirzoian (changes requested):

bogemna (commented):
add

bogemna (commented):
add

src/styles/components/main.scss
```diff
@@ -0,0 +1,19 @@
+* {
+  box-sizing: border-box;
+}
+
+.card {
```
roman-mirzoian, line 5: Why are these styles separate from card.scss file?

src/index.html
```diff
@@ -5,9 +5,44 @@
...
+      <h3 class="card__name">APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</h3>
+      <p class="card__code">Product code: 195434</p>
+      <div class="card__reviews reviews">
+        <div class="reviews__stars stars stars--4">
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+        </div>
```
roman-mirzoian, line 30: You can use more semantic tags for stars.

roman-mirzoian (changes requested):
There is quite a bit left to do

src/styles/components/main.scss
bogemna: because it doesn't work right (Idk why)

I tried(

src/images/Stars.png
alexander-ignatow, line 1: Why do you need this file?

src/index.html
```diff
@@ -5,9 +5,44 @@
...
+    <div class="card" data-qa="card">
+      <img
+        src="./images/imac.jpeg"
+        alt="APPLE A1419 iMac 27' Retina 5K Monoblock (MNED2UA/A)"
+        class="card__img"
+      >
+      <h3 class="card__name">APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</h3>
+      <p class="card__code">Product code: 195434</p>
+      <div class="card__reviews reviews">
+        <div class="reviews__stars stars stars--4">
```
alexander-ignatow, line 24: Do you actually need `reviews__stars` class?

src/styles/components/card.scss
```diff
@@ -0,0 +1,66 @@
...
+    color: #fff;
+    background-color: $button-color;
+    border-radius: 5px;
+    border: 1px solid transparent;
+    @include text-bold;
+
+    &:hover {
+      background-color: $main-color;
+      color: $button-color;
+      border-color: $button-color;
```
alexander-ignatow, line 63: No need to change border-color: notice you are setting it to the same value it was :)

src/styles/components/card.scss
```diff
@@ -0,0 +1,66 @@
...
+
+.card {
+  max-width: 200px;
+  display: flex;
+  flex-direction: column;
+  padding: 32px 16px 16px;
+  border: 1px solid $main-color;
+  border-radius: 5px;
+
+  margin: 0;
```
alexander-ignatow, line 12: Dont set margins to BEM blocks

src/styles/components/card.scss
```diff
@@ -0,0 +1,66 @@
...
+  font-family: "Roboto", Helvetica, sans-serif;
```
alexander-ignatow, line 13: `body` is a better place to set `font-family`

src/images/.gitkeep
alexander-ignatow, line 1: .gitkeep files are needed only when you want to put an "empty" folder to the git, so you don't need it here

alexander-ignatow (changes requested):
Good job cleaning up!
Still, some issues remain - let's deal with them!

src/images/Stars.png
bogemna: I don't remember, it was written week ago.........

BudnikOleksii (changes requested):
Great, but some comments are still not fixed👀
Remove the margin from the BEM block "card" and move the main font-family declaration to the body selector instead of the "card" block.

BudnikOleksii (approved):
Nicely done🔥

PR #3383, opened 2023-07-20
add task solution

desevoker:
- [DEMO LINK](https://desevoker.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://desevoker.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,96 @@
+.card {
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 400;
+  color: $color-text-primary;
+
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  width: 200px;
+  padding: 32px 16px 16px;
+  border-radius: 5px;
+  border: 1px solid $color-white-smoke;
+
+  &__image {
+    display: block;
+    width: 160px;
+    height: 134px;
+    margin-bottom: 40px;
+  }
+
+  &__title {
+    font-weight: 500;
+    line-height: 18px;
+
+    margin-bottom: 4px;
+  }
+
+  &__description {
+    font-size: 10px;
+    line-height: 14px;
+    color: $color-text-secondary;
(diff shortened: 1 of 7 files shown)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,97 @@
+.card {
+  font-family: "Roboto", sans-serif;
```
BudnikOleksii, line 2: Checklist:
image

Also, as it's the main font it's better to apply it for the html or body, let's moe it.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,97 @@
...
+    box-sizing: border-box;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: 100%;
+    height: 40px;
+    background-color: $color-accent-blue;
+    border-radius: 5px;
+    border: 1px solid transparent;
+    transition: 0.3s ease;
```
BudnikOleksii, line 89: If you use transition it's better to specify to what properties it should apply

BudnikOleksii (changes requested):
Nicely done🚀
Just one moment that I mentioned in the comments.

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,6 @@
+$color-text-primary: #060b35;
+$color-text-secondary: #616070;
+
+$color-accent-blue: #00acdc;
+$color-white-smoke: #f3f3f3;
+$color-white: #fff;
```
roman-mirzoian, line 6: You should not name variables by the color that is stored there, because the color can change during the development process and you will have to update this not only in the variable definition, but also throughout the code.
Therefore, it is worth naming variables based only on their purpose.

roman-mirzoian (approved):
Well done 👍
But be careful when naming variables and classes. Name them by purpose, not by content.

src/styles/utils/variables.scss
desevoker: Thanks, that makes sense, now I understand.

PR #3396, opened 2023-07-21
Product cards serkrops

serkrops:


src/styles/utils/__normalize.scss
```diff
@@ -0,0 +1,9 @@
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-size: 12px;
+  font-family: Roboto, sans-serif;
```
BudnikOleksii, line 7: Checklist:
image

src/styles/utils/__reset.scss
```diff
@@ -0,0 +1,7 @@
+body,
+img,
+h3,
+p,
+div {
+  margin: 0;
```
BudnikOleksii, line 6: div and img elements don't have margins, so you don't need to reset them.

BudnikOleksii (changes requested):
Nicely done🔥
Left some comments to improve your code, check them.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,94 @@
...
+    line-height: 16px;
+    text-transform: uppercase;
+
+    text-decoration: none;
+    border: 1px solid $blue-accent;
+    border-radius: 5px;
+    background: $blue-accent;
+
+    &:hover {
+      border: 1px solid $blue-accent;
```
roman-mirzoian, line 89: https://github.com/mate-academy/layout_product-cards/blob/master/checklist.md
9. [STYLES] - Don't add new border to the element on hover. Add default transparent border of the same width, and change its color on :hover

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,5 @@
+$main-text-color: #060b35;
+$secondary-text-color: #616070;
+$white-accent: #fff;
+$blue-accent: #00acdc;
```
roman-mirzoian, line 4: You need to fix the variable names, they should show their purpose, not their content. if I decided to change the color, I would have to change it everywhere, not just in the variable definition.

roman-mirzoian (changes requested):

roman-mirzoian (approved):
Well done 👍

PR #3419, opened 2023-07-29
Develop

yuliiaaaaa:


src/styles/card.scss
```diff
@@ -0,0 +1,93 @@
...
+    background-color: $card-button-color;
+    margin: 16px;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    border: 1px solid transparent;
+    box-sizing: border-box;
+  }
+
+  &__link:hover {
```
denys-danyliuk, line 88: you should put this inside link class, than it will be &:hover

src/styles/card.scss
```diff
@@ -0,0 +1,93 @@
...
+    margin-inline: 16px;
+    margin-top: 40px;
+    color: $card-title-color;
+  }
+
+  &__subtitle {
+    font-size: 10px;
+    line-height: 14px;
+    color: $card-subtitle-color;
+    margin-inline: 16px 84px;
```
denys-danyliuk, line 27: be consistent with your margins, now you have margin collapse, 10'th point in checklist

src/styles/card.scss
```diff
@@ -0,0 +1,93 @@
...
+    margin-top: 4px;
+  }
+
+  &__rate {
+    display: flex;
+    justify-content: space-between;
+    margin-top: 16px;
+    margin-inline: 16px;
```
denys-danyliuk, line 35: same here

src/styles/card.scss
```diff
@@ -0,0 +1,93 @@
+.card {
+  width: $card-width;
+  border-radius: 5px;
+  border: 1px solid $card-border-color;
+  background: $card-background-color;
+  box-sizing: border-box;
```
denys-danyliuk, line 6: try to add paddings to the card instead of adding margin-left/right for elements

src/styles/card.scss
```diff
@@ -0,0 +1,93 @@
...
+
+  &__review {
+    font-size: 10px;
+    line-height: 14px;
+    text-align: right;
+    color: $card-title-color;
+    margin: auto 0;
+  }
+
+  &__priceText {
```
denys-danyliuk, line 46:
```suggestion
  &__price-text {
```
follow bem naming conventions

src/styles/stars.scss
```diff
@@ -0,0 +1,23 @@
...
+    background-position: center;
+
+    &:last-child {
+      margin-right: 0;
+    }
+    .stars--1 &:nth-child(1),
+    .stars--5 &:not(:nth-child(n + 6)),
+    .stars--4 &:not(:nth-child(n + 5)),
+    .stars--3 &:not(:nth-child(n + 4)),
+    .stars--2 &:not(:nth-child(n + 3)) {
```
denys-danyliuk, line 19: use sass nesting here also

denys-danyliuk (changes requested):

src/index.html
```diff
@@ -1,13 +1,48 @@
...
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <img
+        src="./images/imac.jpeg"
+        alt="APPLE A1419 iMAC 27"
+        class="card__img"
+      >
+      <h2 class="card__title">APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</h2>
```
Esceype, line 19: How do think what the font-weight you have on this text ?

src/index.html
```diff
@@ -1,13 +1,48 @@
...
+
+      <div class="card__rate">
+        <div class="stars stars--4">
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+        </div>
+        <p class="card__review">Reviews: 5</p>
```
Esceype, line 30: You have here some margins that don`t need
image

src/styles/card.scss
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  &__link {
+    text-transform: uppercase;
+    font-size: 14px;
+    line-height: 16px;
+    font-weight: 700;
+    text-decoration: none;
+    color: $card-link-color;
+    width: $card-button-width;
```
Esceype, line 71: You don`t need it here as your button always should be 100% of width, obviously given block paddings

src/styles/normalize.scss
```diff
@@ -0,0 +1,17 @@
...
+  margin: 0;
+}
+
+h2 {
+  margin: 0;
+}
+
+p {
+  margin: 0;
+}
```
Esceype, line 17: Try to remove it and than run tests, also try to move styles from html selector to body

src/styles/card.scss
```diff
@@ -0,0 +1,88 @@
...
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    border: 1px solid transparent;
+    box-sizing: border-box;
+
+    &:hover {
+      color: $card-button-color;
+      background-color: $card-link-color;
+      border-color: $card-button-color;
```
Esceype, line 85: Border color don`t change so move this color to static state

Esceype (changes requested):
GJ!!! But let`s try to improve your code and pass tests with come small fixes

src/index.html
yuliiaaaaa: In the mockup the font-weight is 500,but when i assigned this,appearance wasn't the same as in mockup

src/styles/normalize.scss
yuliiaaaaa: when i remove this,i have extra margin at the bottom,but i only assigned margin-top.I think the problem is in default margin,so i reseted this

yuliiaaaaa: Can you explain the difference between assigning common styles to body and to html tags,please?

src/styles/normalize.scss
```diff
@@ -0,0 +1,14 @@
...
+  font-weight: 400;
+}
+
+// h2 {
+//   margin: 0;
+// }
+
+// p {
+//   margin: 0;
+// }
```
xiimiik, line 14: It was a good idea, why did you comment on it?

src/styles/card.scss
xiimiik: not fixed

xiimiik (changes requested):
these elements are clearly larger than necessary
image
you set margin-top everywhere, but here there is a bottom
image

src/styles/card.scss
yuliiaaaaa: But i have padding-inline: 16px; or you mean something else?

xiimiik (approved):
Nice work!

PR #3433, opened 2023-08-01
add task solution

vinogradova8:
- [DEMO LINK](https://vinogradova8.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://vinogradova8.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -3,11 +3,49 @@
   <head>
     <meta charset="UTF-8">
     <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Product cards</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+    <link rel="stylesheet" href="./styles/index.scss">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <img
+        src="images/imac.jpeg"
+        alt="APPLE A1419 iMac 27 Retina 5K Monoblock (MNED2UA/A)"
+        class="card__image"
+      >
+
+      <h2 class="card__title">APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </h2>
+
+      <div class="card__code">Product code: 195434</div>
+
+      <div class="card__rating">
+        <div class="stars stars--4">
+          <span class="stars__star"></span>
+          <span class="stars__star"></span>
+          <span class="stars__star"></span>
+          <span class="stars__star"></span>
+          <span class="stars__star"></span>
+        </div>
+
+        <span class="card__reviews">Reviews: 5</span>
+      </div>
+
+      <div class="card__price">
+        <span class="card__price-name">Price:</span>
+        <span class="card__price-value">$2,199</span>
+      </div>
+
+      <a
+        href="#buy"
+        class="card__button"
+        data-qa="hover">
+        Buy
+      </a>
+    </div>
   </body>
 </html>
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,86 @@
+.card {
+  font-family: "Roboto", "Helvetica", sans-serif;
+  font-size: 12px;
+  line-height: 1.5;
+  box-sizing: border-box;
+  max-width: 200px;
+  padding: 32px 16px 16px;
+  border-radius: 5px;
+  border: 1px solid #f3f3f3;
+  background: #fff;
+  color: $main-text-color;
+
+  &__image {
+    display: block;
+    margin: 0 auto 40px;
+    width: 160px;
+    height: 134px;
+  }
+
+  &__title {
+    font-size: 12px;
+    font-weight: 500;
+    margin-bottom: 4px;
+  }
+
+  &__code {
+    font-size: 10px;
+    margin-bottom: 16px;
+    color: $second-text-color;
+    line-height: 1.4;
+  }
+
+  &__rating {
+    display: flex;
+    align-items: flex-end;
+    justify-content: space-between;
+    margin-bottom: 24px;
+  }
+
+  &__reviews {
+    font-size: 10px;
+    line-height: 1.4;
+  }
+
+  &__price {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    margin-bottom: 16px;
+  }
+
+  &__price-name {
+    color: $second-text-color;
+  }
+
+  &__price-value {
+    font-weight: 700;
+    line-height: 1.12;
+    font-size: 16px;
+  }
+
+  &__button {
+    box-sizing: border-box;
+    background-color: #00acdc;
+    text-align: center;
+    text-decoration: none;
+    padding: 12px 0;
+    border-radius: 5px;
+    color: #fff;
+    font-size: 14px;
+    font-weight: 700;
+    line-height: 1.14;
+    text-transform: uppercase;
+    width: 100%;
+    display: inline-block;
+    border: 1px solid transparent;
+    transition: all 0.3s ease 0s;
+
+    &:hover {
+      background-color: #fff;
+      color: #00acdc;
+      border: 1px solid #00acdc;
+      transition: all 0.3s ease 0s;
+    }
+  }
+}
--- a/src/styles/blocks/stars.scss
+++ b/src/styles/blocks/stars.scss
@@ -1,0 +1,43 @@
+.stars {
+  display: flex;
+
+  &__star {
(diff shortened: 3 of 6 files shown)


src/index.html
```diff
@@ -3,11 +3,49 @@
...
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <img
+        src="images/imac.jpeg"
+        alt="APPLE A1419 iMac 27 Retina 5K Monoblock (MNED2UA/A)"
+        class="card__image"
+      >
+
+      <h2 class="card__title">APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </h2>
```
TarasHoliuk, line 22:
```suggestion
      <h2 class="card__title">
        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
      </h2>
```

src/index.html
```diff
@@ -3,11 +3,49 @@
...
+
+      <div class="card__price">
+        <span class="card__price-name">Price:</span>
+        <span class="card__price-value">$2,199</span>
+      </div>
+
+      <a
+        href="#buy"
+        class="card__button"
+        data-qa="hover">
```
TarasHoliuk, line 46:
```suggestion
      <a
        href="#buy"
        class="card__button"
        data-qa="hover"
      >
```
[checklist](https://github.com/mate-academy/layout_product-cards/blob/master/checklist.md) 14:
image

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,86 @@
+.card {
+  font-family: "Roboto", "Helvetica", sans-serif;
+  font-size: 12px;
+  line-height: 1.5;
+  box-sizing: border-box;
```
TarasHoliuk, line 5: it should be applied to each element on the page (`*`)

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,86 @@
...
+  max-width: 200px;
+  padding: 32px 16px 16px;
```
TarasHoliuk, line 7:
```suggestion
  padding: 32px 16px;
```
the same

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,43 @@
...
+
+    &:not(:first-child) {
+      margin-left: 4px;
+    }
+  }
+}
+
+.stars--1 {
+  .stars__star:first-child {
+    background: url("../images/star-active.svg") center no-repeat;
```
TarasHoliuk, line 17:
```suggestion
    background-image: url("../images/star-active.svg");
```
don't duplicate other background properties

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,43 @@
...
+    width: 16px;
+    background: url("../images/star.svg") center no-repeat;
+
+    &:not(:first-child) {
+      margin-left: 4px;
+    }
+  }
+}
+
+.stars--1 {
```
TarasHoliuk, line 15: use scss loop to avoid duplicating modificator styles

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,86 @@
...
+    text-transform: uppercase;
+    width: 100%;
+    display: inline-block;
+    border: 1px solid transparent;
+    transition: all 0.3s ease 0s;
+
+    &:hover {
+      background-color: #fff;
+      color: #00acdc;
+      border: 1px solid #00acdc;
```
TarasHoliuk, line 82: change border-color only

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,86 @@
...
+      transition: all 0.3s ease 0s;
```
TarasHoliuk, line 83: redundant, defining it above is enough

src/styles/blocks/card.scss
TarasHoliuk: #00acdc is reusable color - move to a variable too

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,86 @@
...
+    border-radius: 5px;
+    color: #fff;
+    font-size: 14px;
+    font-weight: 700;
+    line-height: 1.14;
+    text-transform: uppercase;
+    width: 100%;
+    display: inline-block;
+    border: 1px solid transparent;
+    transition: all 0.3s ease 0s;
```
TarasHoliuk, line 77: it's better to list properties you change instead of `all`

TarasHoliuk (changes requested):

src/index.html
```diff
@@ -3,11 +3,51 @@
...
+          <span class="stars__star"></span>
+          <span class="stars__star"></span>
+        </div>
+
+        <span class="card__reviews">Reviews: 5</span>
+      </div>
+
+      <div class="card__price">
+        <span class="card__price-name">Price:</span>
+        <span class="card__price-value">$2,199</span>
```
denys-danyliuk, line 41: use p instead, 12'th point in checklist

src/index.html
```diff
@@ -3,11 +3,51 @@
...
+        src="images/imac.jpeg"
+        alt="APPLE A1419 iMac 27 Retina 5K Monoblock (MNED2UA/A)"
+        class="card__image"
+      >
+
+      <h2 class="card__title">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </h2>
+
+      <div class="card__code">Product code: 195434</div>
```
denys-danyliuk, line 25: use p

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,88 @@
...
+    border-radius: 5px;
+    color: #fff;
+    font-size: 14px;
+    font-weight: 700;
+    line-height: 1.14;
+    text-transform: uppercase;
+    width: 100%;
+    display: inline-block;
+    border: 1px solid transparent;
+    transition: background-color 0.3s ease 0s, color 0.3s ease 0s;
```
denys-danyliuk, line 80: add border-color also

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,88 @@
+* {
+  font-family: "Roboto", "Helvetica", sans-serif;
+  font-size: 12px;
+  line-height: 1.5;
```
denys-danyliuk, line 4: you can move those to body

denys-danyliuk (changes requested):

src/styles/blocks/card.scss
vinogradova8: Yes, I can. But what for? I also can move these properties to html, there more than one correct versions of code. Moving these properties to another selector will make my code longer. 
Mentor TarasHoliuk left another comment below this part of my code: "it should be applied to each element on the page (*)"

src/index.html
xiimiik: because the `<span>` tag does not carry any semantic content it uses only for styling a parts of text
https://developer.mozilla.org/ru/docs/Web/HTML/Element/span

src/styles/blocks/card.scss
xiimiik: That's why you can't use `*` before anything except `box-sizing: border-box'

Wide range of applications: The `*` selector selects all elements on the page. This means that any styles applied to the * selector will be applied to all elements, including inline elements such as `<script>`, `<style>`, and others. This can lead to unexpected problems and affect the behaviour of scripts and styles.

Impact on performance: Selecting all the elements on a page can lead to an increase in the amount of styles passed to the browser, which can lead to a decrease in site performance. The `*` selector can make styling slower, especially if you apply complex styles or increase the number of elements to be processed.

Overwrite styles: The `*` selector has the highest level of specificity, which means that it overrides (overwrites) other styles, including styles defined with classes or identifiers. This can lead to style conflicts and make it difficult to manage styles on a page.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,94 @@
...
+  box-sizing: border-box;
+}
+
+p {
+  margin: 0;
+}
```
xiimiik, line 10: put it to index.scss, where u already reset h2 and body

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,94 @@
...
+    font-size: 14px;
+    font-weight: 700;
+    line-height: 1.14;
+    text-transform: uppercase;
+    width: 100%;
+    display: inline-block;
+    border: 1px solid transparent;
+    transition: background-color 0.3s ease 0s,
+      color 0.3s ease 0s,
+      border-color 0.3s ease 0s;
```
xiimiik, line 86: move 0.3s to variable

xiimiik (changes requested):

src/styles/blocks/card.scss
vinogradova8: I don`t agree with the last paragraph. The * selector doesn`t impact the specificity weight value.
https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

src/index.html
```diff
@@ -3,11 +3,51 @@
...
+      <div class="card__rating">
+        <div class="stars stars--4">
+          <span class="stars__star"></span>
+          <span class="stars__star"></span>
+          <span class="stars__star"></span>
+          <span class="stars__star"></span>
+          <span class="stars__star"></span>
+        </div>
+
+        <span class="card__reviews">Reviews: 5</span>
```
denys-danyliuk, line 36: turn it into p

denys-danyliuk (approved):

PR #3449, opened 2023-08-04
add task solution

shchipina:
[DEMO LINK](https://shchipina.github.io/layout_product-cards/)
[TEST REPORT LINK](https://shchipina.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -4,10 +4,45 @@
     <meta charset="UTF-8">
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
     <title>Product cards</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="./styles/index.scss">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <section class="content">
+      <div class="card" data-qa="card">
+
+        <img class="card__img" src="images/imac.jpeg"
+          alt="imac"
+        >
+
+        <div class="card__info">
+          <h2 class="card__info-title">APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</h2>
+          <p class="card__info-code">Product code: 195434</p>
+
+          <div class="card__info-stars">
+            <div class="stars stars--4">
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -4,10 +4,45 @@
...
+              <div class="stars__star"></div>
+              <div class="stars__star"></div>
+              <div class="stars__star"></div>
+            </div>
+
+            <p class="card__info-reviews">Reviews: 5</p>
+          </div>
+          <div class="card__price">
+            <p class="card__price-text">Price:</p>
+            <h1 class="card__price-num">$2,199</h1>
```
polosanya, line 36: `h1` should be only 1 per page. Better use h3/h4 instead

src/styles/card.scss
```diff
@@ -0,0 +1,77 @@
+.card {
+  box-sizing: border-box;
+  width: 200px;
```
polosanya, line 3: It would be great to make it a variable. Same with all colors

src/styles/card.scss
```diff
@@ -0,0 +1,77 @@
...
+    padding: 12px 70px;
+    border-radius: 5px;
+  }
+
+  &__buy:hover {
+    background-color: #fff;
+    color: #00acdc;
+    border: 1px solid #00acdc;
+    padding: 11px 69px;
+    transition: 0.4s ease-out;
```
polosanya, line 75: Transition should be added for the default state of the element, not hover, so it works smoothly both ways.
Also, better specify the property you want to apply a transition to
https://stackoverflow.com/questions/8947441/css3-transitions-is-transition-all-slower-than-transition-x

src/styles/stars.scss
```diff
@@ -0,0 +1,25 @@
+.card__info-stars {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
```
polosanya, line 5: It should be inside card block, not stars

src/styles/stars.scss
```diff
@@ -0,0 +1,25 @@
...
+    background-repeat: no-repeat;
+    height: 16px;
+    width: 16px;
+  }
+  &__star:not(:last-child) {
+    margin-right: 4px;
+  }
+
+  &--4 &__star:nth-child(-n+4) {
+  background-image: url(../images/star-active.svg);
```
polosanya, line 21: What about other modifiers? It should work correctly with them too
image

StasSokolov (approved):
Great job! I'm approving it🦾

PR #3453, opened 2023-08-06
add product cards

IwonaIvona:
https://iwonaivona.github.io/layout_product-cards/
https://iwonaivona.github.io/layout_product-cards/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,66 @@
 <!doctype html>
 <html lang="en">
   <head>
-    <meta charset="UTF-8">
-    <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta charset="UTF-8" >     
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Product cards</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="./styles/index.scss">
+    <link 
+      rel="stylesheet"
+      href="https://fonts.googleapis.com/css2f?family=Roboto:wght@400;500;700&display=swap"
+    >
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <img
+        src="../src/images/imac.jpeg"
+        alt="APPLE A1419 iMac 27 Retina 5K Monoblock (MNED2UA/A)"
+        class="card__img"
+      >
+
+      <h1 class="card__product">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </h1>
+
+      <p class="card__code">
+        Product code: 195434
+      </p>
+
+      <div class="card__review">
+        <div class="stars stars--4">
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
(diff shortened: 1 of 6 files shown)


IwonaIvona:
Sorry for posting the wrong answer. I had a problem with the correct display of the tests. I will correct it.

IwonaIvona:
I have a problem with setting the font weight in the text. According to Figma, I have the correct one, but it displays differently.

src/styles/variables.scss
```diff
@@ -0,0 +1,5 @@
+:root {
+  --main-accent: #060b35;
+  --secondary-accent: #616070;
```
darokrk, line 3: You're using CSS variables instead of SCSS variables. In SCSS variables should be defined at the top of the SCSS file using $ sign.

src/styles/variables.scss
```diff
@@ -0,0 +1,5 @@
+:root {
```
darokrk, line 1: You're using CSS variables instead of SCSS variables. In SCSS variables are defined using $ sign.

src/styles/variables.scss
```diff
@@ -0,0 +1,5 @@
...
+  --main-accent: #060b35;
+  --secondary-accent: #616070;
+  --blue-accent: #00acdc;
```
darokrk, line 4: You're using CSS variables instead of SCSS variables. In SCSS variables should be defined at the top of the SCSS file using $ sign.

src/styles/variables.scss
```diff
@@ -0,0 +1,5 @@
+:root {
+  --main-accent: #060b35;
```
darokrk, line 2: You're using CSS variables instead of SCSS variables. In SCSS variables should be defined at the top of the SCSS file using $ sign.

src/index.html
```diff
@@ -1,13 +1,63 @@
...
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+    <link rel="stylesheet" href="./styles/index.scss">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <img
+        src="../src/images/imac.jpeg"
```
darokrk, line 17: The path to the image seems to be incorrect. You are currently in the src directory, so you don't need to go back to it. You can use 'images/imac.jpeg' instead.

src/styles/card.scss
```diff
@@ -0,0 +1,93 @@
+.card {
+  padding: 32px 16px 16px;
+  display: flex;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  box-sizing: border-box;
```
darokrk, line 6: It's a good practice to set box-sizing to border-box globally for all elements at the root level of your styles, to ensure consistent layout calculations across all elements. Consider moving this to a global stylesheet to index.scss

darokrk (changes requested):
UI looks good 👍🏽  add a few mentions to code improvement 🛠️ , also test are failing, please fix them following Figma's design 😃  seems spacing or font sizes are not the same:

image

src/index.html
IwonaIvona: When I change according to your directions, I get Error during test :)

darokrk: ok this can be omit :)

src/styles/card.scss
```diff
@@ -0,0 +1,93 @@
...
+  background: #fff;
```
darokrk, line 7: this color can be stored in variable, bcs you are using this multiple times

src/styles/card.scss
darokrk: please move this `index.scss` file and use it for every element:

```
* {
  box-sizing: border-box;
}
```

darokrk (changes requested):

darokrk (approved):
Good job 🔥 🏆 !

PR #3500, opened 2023-08-18
add task solution

Daniilart01:
I have a little issue with tests, in my implementation I am using `max-width: 100%` to fit mac image into card, but in test it's 1px bigger and I have no idea why does it happens (as I see, my image shrinks on the y-axis)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -5,9 +5,46 @@
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Product cards</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="./styles/index.scss">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <div class="card__image-container">
+        <img
+        src="images/imac.jpeg"
+        alt="APPLE A1419 iMac 27&quot; Retina 5K Monoblock (MNED2UA/A)"
+        class="image-product"
+      >
+      </div>
+      <div class="info">
+        <h2 class="info__title">
+          APPLE A1419 iMac 27&quot; Retina 5K Monoblock (MNED2UA/A)
+        </h2>
+
+        <p class="code-label">Product code: 195434</p>
+        <div class="info__reviews">
+          <div class="stars stars--4">
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+          </div>
+
+          <p class="reviews-label">Reviews: 5</p>
+        </div>
+        <div class="info__price">
+          <span class="price-label">Price:</span>
+          <p class="price-value">&#36;2,199</p>
+        </div>
+
+        <div class="info__buy" data-qa="hover">
+          <a href="#" class="buy-link">Buy</a>
+        </div>
+      </div>
+    </div>
   </body>
 </html>
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,117 @@
+* {
+  --main-accent: #060b35;
+  --secondary: #616070;
+  --blue-accent: #00acdc;
+}
+
+.card {
+  font-family: "Roboto", "Arial", sans-serif;
+  width: 200px;
+  box-sizing: border-box;
+  border-radius: 5px;
+  border: 1px solid #f3f3f3;
+  background-color: #fff;
+}
+
+.card__image-container {
+  display: flex;
+  padding: 32px 19px 0;
+}
+
+.image-product {
+  max-width: 100%;
+  max-height: 100%;
+}
+
+.info {
+  padding: 40px 16px 16px;
+}
+
+.info__title {
+  margin: 0;
+  color: var(--main-accent);
+  font-size: 12px;
+  font-weight: 500;
+  line-height: 150%;
+}
+
+.code-label {
+  margin: 4px 0 0;
+  color: var(--secondary);
+  font-size: 10px;
+  font-weight: 400;
+  line-height: 140%;
+}
+
+.info__reviews {
+  margin-top: 16px;
+  display: flex;
+  justify-content: space-between;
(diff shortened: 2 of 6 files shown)


src/index.html
```diff
@@ -5,9 +5,46 @@
...
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <div class="card__image-container">
+        <img
+        src="images/imac.jpeg"
+        alt="APPLE A1419 iMac 27&quot; Retina 5K Monoblock (MNED2UA/A)"
+        class="image-product"
+      >
```
Andriy-Shymkiv, line 20:
```suggestion
        <img
          src="images/imac.jpeg"
          alt="APPLE A1419 iMac 27&quot; Retina 5K Monoblock (MNED2UA/A)"
          class="image-product"
        >
```

src/index.html
```diff
@@ -5,9 +5,46 @@
...
+      </div>
+      <div class="info">
```
Andriy-Shymkiv, line 22: i think you can get rid of this wrapper and just give padding's to card block

src/index.html
```diff
@@ -5,9 +5,46 @@
...
+        <h2 class="info__title">
+          APPLE A1419 iMac 27&quot; Retina 5K Monoblock (MNED2UA/A)
+        </h2>
+
+        <p class="code-label">Product code: 195434</p>
+        <div class="info__reviews">
```
Andriy-Shymkiv, line 28:
```suggestion
        <p class="code-label">Product code: 195434</p>
        
        <div class="info__reviews">
```

src/index.html
```diff
@@ -5,9 +5,46 @@
...
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+          </div>
+
+          <p class="reviews-label">Reviews: 5</p>
+        </div>
+        <div class="info__price">
```
Andriy-Shymkiv, line 39:
```suggestion
        </div>
        
        <div class="info__price">
```

src/index.html
```diff
@@ -5,9 +5,46 @@
...
+          <span class="price-label">Price:</span>
```
Andriy-Shymkiv, line 40: use semantic

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,117 @@
...
+  --blue-accent: #00acdc;
+}
+
+.card {
+  font-family: "Roboto", "Arial", sans-serif;
+  width: 200px;
+  box-sizing: border-box;
+  border-radius: 5px;
+  border: 1px solid #f3f3f3;
+  background-color: #fff;
```
Andriy-Shymkiv, line 13: it is white by default

Andriy-Shymkiv:
> I have a little issue with tests, in my implementation I am using `max-width: 100%` to fit mac image into card, but in test it's 1px bigger and I have no idea why does it happens (as I see, my image shrinks on the y-axis). P.S I realized that if i set height to image container manualy (134px) test passed.

then just do:
```
 max-height: 134px;
height: 100%;
```

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,117 @@
...
+
+.info__reviews {
+  margin-top: 16px;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
+
+.reviews-label {
+  margin: 0;
```
Andriy-Shymkiv, line 54: reset margins for each element you need at once

Andriy-Shymkiv (changes requested):

src/index.html
Daniilart01: We have different paddings for image-block and for info-block, so i decide to create to separate containers for image and for info instead of adding additional paddings for image

Andriy-Shymkiv: make sense, agree

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,115 @@
...
+
+.card {
+  font-family: "Roboto", "Arial", sans-serif;
+  width: 200px;
+  box-sizing: border-box;
+  border-radius: 5px;
+  border: 1px solid #f3f3f3;
+}
+
+.card__image-container {
```
Andriy-Shymkiv, line 15: please use scss nesting here and everywhere possible.

Andriy-Shymkiv (changes requested):
good job! you fixed everything from previous review, but one thing that I forget - scss nesting. It is a very good practice to use it, so please add it wherever possible

src/styles/blocks/card.scss
Daniilart01: +

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,115 @@
...
+    height: 40px;
+    background: var(--blue-accent);
+
+    &:hover {
+      background-color: #fff;
+
+      &-link {
+        color: var(--blue-accent);
+      }
+    }
```
Esceype, line 97: On hover should change the color of text but this doesn`t work you can make it without this wrapper &__buy and without display: flex, just use display block and width: 100% and paddings and then centralize text.

src/styles/index.scss
```diff
@@ -1,3 +1,9 @@
 body {
   margin: 0;
 }
+
+p {
+  margin: 0;
+}
```
Esceype, line 7:
```suggestion
body,
p {
  margin: 0;
}
```

Esceype (changes requested):
Almost done but don`t forget to add demo link and test report link cause at the moment your hover effect doesn`t work as expected

src/styles/blocks/card.scss
Daniilart01: yeah, my bad, the reason is in `&` symbol, cause when i am applying it in row 94 in css it appears as `.info__buy:hover-link`, but we need `.info__buy:hover .info__buy-link`,

Daniilart01: this fixed by changing ` &-link` to  `& .info__buy-link`

Daniilart01:
I replaced `display: flex` to `display: block; text-align: center; line-height: 38px`. `Line-hight:38px` because we have `border-box` with `height: 40px` and `1px` borders. But i still think it's wrong because we should set up hardcoded value `38px` (even not 40 because of `border-box`). For other developers it may be hard to understand why exactly `38px`, so I guess, flex block with align-items and justify-content is better option

Daniilart01:
I replaced display: flex to display: block; text-align: center; line-height: 38px. Line-hight:38px because we have border-box with height: 40px and 1px borders. But i still think it's wrong because we should set up hardcoded value 38px (even not 40 because of border-box). For other developers it may be hard to understand why exactly 38px, so I guess, flex block with align-items and justify-content is better option

mykhalenych (approved):

PR #3503, opened 2023-08-19
add task solution

pzwolenik:
- [DEMO LINK](https://pzwolenik.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://pzwolenik.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,28 +1,4 @@
-// @use "./blocks/card.scss";
-// @use "./blocks/stars.scss";
 // @use "../utils/variables" as *;
-// For some reason this didn't work AT ALL,
-// so sorry, I'm sticking with all code in one SCSS
-
-$main-accent: #060b35;
-$secondary: #616070;
-$white: #fff;
-$grey-elements: #f3f3f3;
-$blue-accent: #00acdc;
-$font-regular: 400;
-$font-medium: 500;
-$font-bold: 700;
-
-:root {
-  font-family: Roboto, Helvetica, sans-serif;
-  font-weight: $font-regular;
-  font-size: 12px;
-  line-height: 18px;
-}
-
-body {
-  margin: 0;
-}
 
 .card {
   box-sizing: border-box;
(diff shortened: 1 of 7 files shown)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,91 @@
+// @use "../utils/variables" as *;
```
OleziO, line 1: Remove all comments from the code.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,91 @@
...
+
+.card {
+  box-sizing: border-box;
```
OleziO, line 4: Move to * selector and remove from all other places.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,91 @@
...
+    font-size: 16px;
+    font-weight: $font-bold;
+  }
+
+  &__button {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+
+    box-sizing: border-box;
```
OleziO, line 72: Same as previous situation

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,25 @@
+// @use "../utils/variables" as *;
```
OleziO, line 1: Remove comment

src/styles/index.scss
```diff
@@ -1,3 +1,139 @@
+// @use "./blocks/card.scss";
```
OleziO, line 1: Remove all comments.

OleziO (changes requested):
Move hover only on button, not a card.
image

pzwolenik:
Hi! Thank you for your time, but I have few quick questions :) (I wrote this few days ago, but I'm pretty sure you didn't get the message.)

1. Can you explain comment a little more? What do you mean by '_Move to * selector_'?
2. Possibly, you can give me an answer, why my code didn't work AT ALL, when I apply @use in index.scss?

It would be very helpful, thanks :)

DorotaLeniecDev:
@pzwolenik yes in your working setup @use syntax will not work due to the fact the project is using node-sass https://github.com/mate-academy/layout_product-cards/blob/master/package.json#L32 which does not support this syntax. You need to use @import syntax

src/styles/blocks/card.scss
DorotaLeniecDev: AD. 1 - You could use universal selector to set `box-sizing: border-box` like so:
```css
*,
*:before,
*:after {
  box-sizing: border-box;
}
```

And you will no longer need to set it to a single element

src/styles/index.scss
```diff
@@ -1,3 +1,137 @@
...
+@use "./utils/variables";
+
+$main-accent: #060b35;
+$secondary: #616070;
+$white: #fff;
+$grey-elements: #f3f3f3;
+$blue-accent: #00acdc;
+$font-regular: 400;
+$font-medium: 500;
+$font-bold: 700;
```
DorotaLeniecDev, line 12: Why you are declaring variables here and in /utils/variables? Do it only one time to avoid confusion.

Alos there is no need for $font-medium/bold/regular variable. It is more readable to put it directly in `font-weight` property

DorotaLeniecDev (changes requested):
Check my comments

carlos-cne (approved):
Your code looks good and works fine.

PR #3519, opened 2023-08-24
add task solution card

Kuper18:
Hi! Here is my card solution.

demo link for the page https://kuper18.github.io/layout_product-cards/


Kuper18 (commented):
Hi! Here is the updated code, thanks for advice

StasSokolov (approved):
Great job🔥 I'm approving it!

PR #3525, opened 2023-08-26
add task solution

yuliiababych:
[DEMO LINK](https://yuliiababych.github.io/layout_product-cards/)
[TEST REPORT LINK](https://yuliiababych.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/styles/card-content.scss
+++ b/src/styles/card-content.scss
@@ -1,0 +1,111 @@
+$main-text-color: #060b35;
+$secondary-text-color: #616070;
+
+html {
+  font-family: Roboto, Helvetica, sans-serif;
+}
+
+body {
+  width: 200px;
+}
+
+.card {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  flex-direction: column;
+  padding: 17px;
+
+  &__image {
+    width: 160px;
+    height: 134px;
+    flex-shrink: 0;
+    margin: 16px 0 40px;
+  }
+}
+
+.card-content {
+  display: flex;
+  justify-content: center;
+  flex-direction: column;
+
+  &__title {
+    color: $main-text-color;
+    font-size: 12px;
+    font-style: normal;
+    font-weight: 500;
+    line-height: 18px;
+    text-align: left;
+    margin: 0;
+  }
+
+  &__code {
+    color: $secondary-text-color;
+    font-size: 10px;
+    font-style: normal;
+    font-weight: 400;
+    line-height: 14px;
+    text-align: left;
+    margin: 4px 0 17px;
+  }
+
+  &__price {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    margin-top: 7px;
+
+    &-label {
+      color: $secondary-text-color;
+      font-size: 12px;
+      font-style: normal;
+      font-weight: 400;
+      line-height: 18px;
+    }
+
+    &-value {
+      color: $main-text-color;
+      text-align: right;
+      font-size: 16px;
+      font-style: normal;
+      font-weight: 700;
+      line-height: 18px;
+    }
+  }
+
+  &__button {
+    display: block;
+    box-sizing: border-box;
+    position: relative;
+    align-self: center;
+    margin: 0;
+    width: 166px;
+    height: 40px;
+    flex-shrink: 0;
+    border-radius: 5px;
+    background-color: #00acdc;
+
+    &-buy {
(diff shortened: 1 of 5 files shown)


src/styles/card-stars.scss
```diff
@@ -0,0 +1,46 @@
+.card-content__stars {
```
TarasHoliuk, line 1: it's a BEM-mix, other words it's a block that at the same time **is an element** of another block
So it should live in the `card-content.scss` together with other elements of the `card-content` block

src/styles/card-stars.scss
```diff
@@ -0,0 +1,46 @@
...
+  justify-content: space-between;
+  align-items: center;
+  margin: 0;
+
+  &-review {
+    color: $main-text-color;
+    text-align: right;
+    font-size: 10px;
+    font-style: normal;
+    font-weight: 400;
```
TarasHoliuk, line 12: default values - no need to set them

src/styles/card-content.scss
```diff
@@ -0,0 +1,111 @@
...
+      text-align: center;
+      text-decoration: none;
+      color: #fff;
+      font-size: 14px;
+      font-style: normal;
+      font-weight: 700;
+      line-height: 16px;
+      text-transform: uppercase;
+
+      &:hover {
```
TarasHoliuk, line 102: to see text I should hover exactly the text:
[Screencast from 27-08-23 10:35:37.webm](https://github.com/mate-academy/layout_product-cards/assets/86923089/39497535-f5a9-4350-8e80-9005c8be1e18)


it should be visible on the button hover

generally, you don't need `button` div wrapper - just apply all styles you need to `a` tag (but of course using the class selector)

src/styles/card-content.scss
```diff
@@ -0,0 +1,111 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  flex-direction: column;
+  padding: 17px;
+
+  &__image {
+    width: 160px;
+    height: 134px;
+    flex-shrink: 0;
```
TarasHoliuk, line 22: redundant

src/styles/card-content.scss
```diff
@@ -0,0 +1,111 @@
+$main-text-color: #060b35;
+$secondary-text-color: #616070;
+
+html {
+  font-family: Roboto, Helvetica, sans-serif;
+}
+
+body {
+  width: 200px;
```
TarasHoliuk, line 9: set it for `card`, not for `body`

src/styles/card-content.scss
```diff
@@ -0,0 +1,111 @@
+$main-text-color: #060b35;
+$secondary-text-color: #616070;
+
+html {
+  font-family: Roboto, Helvetica, sans-serif;
+}
```
TarasHoliuk, line 6: shouldn't be here, since it's not part of `card` block

src/styles/card-content.scss
```diff
@@ -0,0 +1,111 @@
...
+    display: block;
+    box-sizing: border-box;
+    position: relative;
+    align-self: center;
+    margin: 0;
+    width: 166px;
+    height: 40px;
+    flex-shrink: 0;
+    border-radius: 5px;
+    background-color: #00acdc;
```
TarasHoliuk, line 86: repeatable color - create a variable for it too

src/styles/card-content.scss
```diff
@@ -0,0 +1,111 @@
...
+      font-size: 16px;
+      font-style: normal;
+      font-weight: 700;
+      line-height: 18px;
+    }
+  }
+
+  &__button {
+    display: block;
+    box-sizing: border-box;
```
TarasHoliuk, line 78: should be in `*` selector only

src/styles/card-content.scss
```diff
@@ -0,0 +1,111 @@
...
+
+  &__image {
+    width: 160px;
+    height: 134px;
+    flex-shrink: 0;
+    margin: 16px 0 40px;
+  }
+}
+
+.card-content {
```
TarasHoliuk, line 27: it seems like you have 2 BEM-blocks here: `card` and `card-content`. There should be only one (`card`). You may just get rid of the `card-content` wrapper and make all its children `card` block's children

TarasHoliuk (changes requested):
GJ :+1: 
Let's improve your code a bit

anastasiiavorobiova (approved):
Awesome!

PR #3556, opened 2023-09-05
add task solution

noraaudonnet:
https://noraaudonnet.github.io/layout_product-cards/
https://noraaudonnet.github.io/layout_product-cards/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -19,9 +19,9 @@
       >
 
       <div class="card__content">
-        <div class="card__name">
+        <h1 class="card__name">
           APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
-        </div>
+        </h1>
 
         <div class="card__code">
           Product code: 195434
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -19,12 +19,12 @@
     font-size: 12px;
     font-weight: 500;
     line-height: 18px;
-    color: #060b35;
+    color: $main-text-color;
     margin-bottom: 4px;
   }
 
   &__code {
-    color: #616070;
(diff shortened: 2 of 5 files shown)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,78 @@
...
+    text-decoration: none;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    box-sizing: border-box;
+    width: 166px;
+    height: 40px;
+    border-radius: 5px;
+    color: #fff;
+    background-color: #00acdc;
```
TarasHoliuk, line 67: let's create variables for repeatable colors

src/index.html
```diff
@@ -5,9 +5,54 @@
...
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <img
+        src="./images/imac.jpeg"
+        alt="Product image"
+        class="card__image"
+      >
+
+      <div class="card__content">
+        <div class="card__name">
```
TarasHoliuk, line 22: use semantic tag for title

src/index.html
```diff
@@ -5,9 +5,54 @@
...
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <img
+        src="./images/imac.jpeg"
+        alt="Product image"
+        class="card__image"
+      >
+
+      <div class="card__content">
```
TarasHoliuk, line 21: I guess you may get rid of this `card__content` wrapper

noraaudonnet: I didn't get rid of it because it makes the right position, although there are no styles for that specifically.
if I remove it, all elements are centered and not spaced between because of "card" properties.

PR #3677, opened 2023-10-11
task solution

bertsu02:
https://bertsu02.github.io/layout_product-cards/
https://bertsu02.github.io/layout_product-cards/report/html_report/


src/index.html
```diff
@@ -5,9 +5,52 @@
...
+      <img
+        src="./images/imac.jpeg"
+        alt="Product image"
+        class="card__image"
+      >
+
+      <div class="card__content">
+        <h1 class="card__name">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </h1>
```
mvjl000, line 24: I think a `p` tag would be better here. H1-H6 tags should represent section headings. Check [the docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements)

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,24 @@
+.stars {
+  display: flex;
+
+  &__star {
+    background-image: url(../images/star.svg);
+    background-position: center;
+    background-repeat: no-repeat;
+    height: 16px;
+    width: 16px;
```
mvjl000, line 9: Variable ;)

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,24 @@
+.stars {
+  display: flex;
+
+  &__star {
+    background-image: url(../images/star.svg);
+    background-position: center;
+    background-repeat: no-repeat;
```
mvjl000, line 7: You could use a shorthand and combine those values in a background property

mvjl000 (changes requested):
Very good job, really! I've left some suggestions to improve your solution ;)

src/index.html
bertsu02: I thougth as it's the header of the card, if I change to p my whole code goes crap. Is it really that much of a problem ?

mvjl000 (changes requested):
Well done! 🚀

mvjl000 (approved):
I gave you request changes accidentaly 😬

PR #3701, opened 2023-10-15
add task solution

K4rol4:


mbruhler (changes requested):
Your links doesn't point to Github Pages so I can't see the final result view. Please check it and correct them to proper ones.

PR #3714, opened 2023-10-18
added two scss files, problem with test

emilionila:
Hi!
I dont get which of css paramaters makes tests go wrong. Hadnt put variables in different file yet.  

- [DEMO LINK](https://emilionila.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://emilionila.github.io/layout_product-cards/report/html_report/)

- [ ] SCSS Variables are used for main values and placed in a **separate file**

### DIFF
--- a/src/styles/card.scss
+++ b/src/styles/card.scss
@@ -1,0 +1,100 @@
+$mainTextColor: #060b35;
+$secondaryTextColor: #616070;
+$whiteColor: #fff;
+$blueAccent: #00acdc;
+
+.card {
+  border: 1px solid #f3f3f3;
+  width: 199px;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding: 32px 0 16px;
+  border-radius: 5px;
+
+  &__img {
+    max-width: 100%;
+    width: 160px;
+    margin-bottom: 40px;
+  }
+
+  &__title {
+    margin: 0 16px;
+    font-weight: 500;
+    font-size: 12px;
+    line-height: 18px;
+    color: $mainTextColor;
+  }
+
+  &__code {
+    margin: 4px 16px 16px;
+    align-self: baseline;
+    font-weight: 400;
+    font-size: 10px;
+    line-height: 14px;
+    color: $secondaryTextColor;
+  }
+
+  &__evaluation {
+    display: flex;
+    justify-content: space-between;
+    align-items: flex-end;
+    width: 100%;
+    margin-bottom: 24px;
+
+    &-review {
+      font-weight: 400;
+      font-size: 10px;
+      line-height: 14px;
+      text-align: right;
+      color: $mainTextColor;
+      padding-right: 16px;
+    }
+  }
+
+  &__price {
+    display: flex;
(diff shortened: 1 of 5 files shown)


src/styles/card.scss
```diff
@@ -0,0 +1,100 @@
...
+  &__title {
+    margin: 0 16px;
+    font-weight: 500;
+    font-size: 12px;
+    line-height: 18px;
+    color: $mainTextColor;
+  }
+
+  &__code {
+    margin: 4px 16px 16px;
```
BudnikOleksii, line 30: To avoid specifying `margin: 0 16px;` to all elements just add padding to the parent element

src/styles/card.scss
```diff
@@ -0,0 +1,100 @@
...
+    font-size: 14px;
+    line-height: 14.55px;
+    text-align: center;
+    color: $whiteColor;
+    background-color: $blueAccent;
+    padding: 12px 70px 12px 69px;
+    text-transform: uppercase;
+    border-radius: 5px;
+    border: 1px solid $blueAccent;
+    transition: all 0.2s ease-in-out;
```
BudnikOleksii, line 91: If you use transition, specify to what properties it should apply

src/styles/index.scss
```diff
@@ -1,3 +1,21 @@
+// @import url(card.scss);
+// @import url(stars.scss);
```
BudnikOleksii, line 2: Use imports, to avoid incorrect styles at first import reset/normalize, variables, mixins, and then other files that use them

src/styles/index.scss
```diff
@@ -1,3 +1,21 @@
...
+
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap");
+
+$bodyFont: "Roboto", sans-serif;
```
BudnikOleksii, line 6: You don't need this variable as it's a default font in your case

src/styles/index.scss
```diff
@@ -1,3 +1,21 @@
...
+}
+
+p {
+  margin: 0;
+}
+
+a {
+  margin: 0;
+  text-decoration: none;
 }
```
BudnikOleksii, line 21: Let's create `reset.scss` file and move to it. Keep your main file clean

BudnikOleksii (changes requested):
Good work 👍 
Left some comments with suggestions. Also, tests are failed

src/styles/card.scss
```diff
@@ -0,0 +1,99 @@
+.card {
+  border: 1px solid #f3f3f3;
+  width: 200px;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding: 32px 0 16px;
+  border-radius: 5px;
+  max-width: 198.4px;
```
emilionila, line 9: I updated code yesterday. All tests work, but i`m not sure that line 9 is correct, without it width of card is 201.6px instead of 200px

emilionila: fixed everything

And678 (approved):
Awesome :fire:

PR #3731, opened 2023-10-19
added product card

AlNeon:
- [DEMO LINK](https://AlNeon.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://AlNeon.github.io/layout_product-cards/report/html_report/)


src/styles/blocks/product-card.scss
```diff
@@ -0,0 +1,83 @@
+.product-card {
+  width: 200px;
+  height: 408px;
```
anastasiiavorobiova, line 3:
```suggestion

```
There is no need to hardcode the height for the card, usually, we have a different amount of content data in product cards

src/styles/blocks/product-card.scss
```diff
@@ -0,0 +1,83 @@
...
+
+    color: #fff;
+    font-size: 14px;
+    font-weight: 700;
+    line-height: 16px;
+    text-transform: uppercase;
+    transition: $std-transition;
+
+    &:hover {
+      background-color: #fff;
```
anastasiiavorobiova, line 78: Consider using variables for repeating colors

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/styles/blocks/product-card.scss
```diff
@@ -0,0 +1,82 @@
+.product-card {
+  width: 200px;
+  background-color: white;
+  border-radius: 5px;
+  border: 1px #f3f3f3 solid;
+  box-sizing: border-box;
+  padding: 33px 16px 16px;
```
etojeDenys, line 7:
```suggestion
  padding: 32px 16px 16px;
```
i suggest it should fix the tests

src/styles/blocks/product-card.scss
```diff
@@ -0,0 +1,82 @@
...
+
+  &__price {
+    margin-top: 24px;
+
+    @include flexSpace;
+
+    &-name {
+      color: $secondary-text-color;
+      font-size: 12px;
+      font-weight: 400;
```
etojeDenys, line 49: it's the default value for `p` tag, so we can easily remove it

etojeDenys (changes requested):
try to fix tests or feel free to ask for some help in the chat

src/styles/blocks/product-card.scss
AlNeon: I deleted the font-size - everything is ok, but if I remove the font-weight, discrepancies with the layout appear. So I left it.

src/styles/blocks/product-card.scss
AlNeon: I changed the picture format from SVG to JPEG and hoped that the tests would pass, but nothing changed. This is where the 1 pixel offset comes from. In the chat mentor told me that with such a minimal difference with the tests, i can leave it as is and do the PR.

volodymyr-soltys97 (approved):
Good job 👍

PR #3742, opened 2023-10-24
Develop

vitalii-labunskyi:


src/index.html
```diff
@@ -5,9 +5,59 @@
...
+          <div class="card__star"></div>
+          <div class="card__star"></div>
+          <div class="card__star"></div>
+          <div class="card__star"></div>
+          <div class="card__star"></div>
+        </div>
+
+        <div class="card__feedbacks">
+          Reviews: 5
+        </div>
```
VitaliyBodnar, line 41: use semantic tags for text

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,131 @@
+.card {
+  display: flex;
+  box-sizing: border-box;
```
VitaliyBodnar, line 3: box-sizing: border-box; is duplicated, so make sense to move it to the `*{ box-sizing: border-box }`

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,131 @@
...
+  flex-direction: column;
+  flex-shrink: 1;
+  width: 200px;
+  height: auto;
+  font-family: "Roboto", sans-serif;
+  border: 1px solid #f3f3ff;
```
VitaliyBodnar, line 9: It is a good practice to make all colors [variable](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties)

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,131 @@
...
+  &__content {
+    margin: 0;
+    font-weight: $font-weight-medium;
+    font-size: $font-size-m;
+    line-height: $line-height-m-xl;
+    color: $theme-content-main;
+    margin-bottom: 4px;
+  }
+
+  &__ID {
```
VitaliyBodnar, line 33: use lower-case for naming id or classes

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,131 @@
...
+    margin: 0;
+    font-style: normal;
```
VitaliyBodnar, line 35: normal is a default value for font-style

src/styles/utils/_reset.scss
```diff
@@ -0,0 +1,5 @@
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
```
VitaliyBodnar, line 4: only box-sizing can be in this block `*{}`

VitaliyBodnar (changes requested):

src/styles/utils/_reset.scss
vitalii-labunskyi: Could you explain, please, why only box-sizing? I want to reset all margins and paddings

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #3933, opened 2023-12-28
add task solution

Shevchuchka:
- [DEMO LINK](https://Shevchuchka.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://Shevchuchka.github.io/layout_product-cards/report/html_report/)


src/index.html
```diff
@@ -2,12 +2,53 @@
...
+        </section>
+        <p class="review__reviews">Reviews: 5</p>
+      </div>
+
+      <div class="price">
+        <p class="price__text">Price:</p>
+        <h1 class="price__number">$2,199</h1>
+      </div>
+
+      <div class="buy">
```
etojeDenys, line 44: we make element a BEM block if we can reuse it somewhere, as an independent block

it seems to me that we will not be able to reuse these elements anywhere, so it is better to make them simple elements of the card block

you must to have only two blocks here: "card" and "stars"

etojeDenys (changes requested):

src/styles/stars.scss
```diff
@@ -0,0 +1,31 @@
...
+  display: flex;
+
+  &__star {
+    margin-right: 4px;
+    width: 16px;
+    height: 16px;
+
+    background-image: url(../images/star.svg);
+    background-repeat: no-repeat;
+    background-position: center;
```
loralevitska, line 11: `background` shorthand can be used here

loralevitska (changes requested):
Hi, I see you haven't fixed previous comments, maybe you did not push your changes? If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

src/index.html
OleziO: Rename class according to the BEM

src/styles/card.scss
```diff
@@ -0,0 +1,9 @@
+.card-block {
+  border: 1px #F3F3F3 solid;
+  border-radius: 5px;
+  box-sizing: border-box;
```
OleziO, line 4: Move to the * selector and remove it from all other places.

OleziO (changes requested):

PR #3976, opened 2024-01-14
Develop

Nastyahumeniuk:
- [DEMO LINK](https://Nastyahumeniuk.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://Nastyahumeniuk.github.io/layout_product-cards/report/html_report/)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,99 @@
...
+
+&__review {
+  display: flex;
+  justify-content: space-between;
+  align-self: center;
+  padding-bottom: 24px;
+
+  &-count {
+    margin: 0;
+    font-family: inherit;
```
TarasHoliuk, line 47: it is inherited by default - no need to explicitly do it every time (only in case it's overwritten by browser default styles - you may check it in devTools in this case):
image

[Article](https://www.w3schools.com/cssref/pr_font_font-family.php)

TarasHoliuk (approved):
Well done :+1:  Approved :heavy_check_mark: 

Added one comment just FYI (no need to re-request review - already approved)

PR #4110, opened 2024-02-07
add task solution

damian2403:
- [DEMO LINK](https://damian2403.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://damian2403.github.io/layout_product-cards/report/html_report/)


src/styles/blocks/variable.scss
```diff
@@ -0,0 +1,2 @@
+$maincolor: #{#00ACDC};
```
asakevych, line 1: name of file should be `_variables.scss`

asakevych (changes requested):
Good job! but some changes are required
remember about deploying your code after push 
check if tests are passed, without tests it can't be accepted

src/styles/blocks/variable.scss
damian2403: src/styles/index.scss
 1:1  ✖  Unexpected leading underscore in imported partial name  scss/load-no-partial-leading-underscore

1 problem (1 error, 0 warnings)

choeqq: @damian2403 you can name it `variables.scss` it doesn't really matters. the point is that you'll have a separate file for variables

choeqq (changes requested):
There are still some color values that should be in variables file, please make sure to move them. 

Also your tests are little bit off, please try to adjust your solution to design

PR #4152, opened 2024-02-21
add task solution

olafchuszno:
- [DEMO LINK](https://olafchuszno.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://olafchuszno.github.io/layout_product-cards/report/html_report/) 

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -20,13 +20,13 @@ Create a pages with product card using `flexbox`, `BEM` and `SCSS` based on [thi
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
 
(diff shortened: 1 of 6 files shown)


yevhenii-pyl (approved):
Good job! 

Lyrical FYI: I'm not a big fan of using margins to position elements inside a container. 
As a rule of thumb, the container should be responsible for its' children positioning.

I've noticed that there is a common pattern with `margin-top: 16px`, why don't you use a `gap` property instead? 
From the other side, there are different margins, like `40px || 24px` and this may become a bit of a nasty problem. In such cases, I prefer to change the html structure adding logical blocks, like: img  => card description => button. 

Instead of setting margin-top for image, why don't you just increase padding-top for a container?

PR #4219, opened 2024-03-06
develop

ropanefarias:
[DEMO LINK](https://ropanefarias.github.io/layout_product-cards/)
[TEST REPORT LINK](https://ropanefarias.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/styles/components/card.scss
+++ b/src/styles/components/card.scss
@@ -1,0 +1,94 @@
+.card {
+  box-sizing: border-box;
+  width: 200px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  padding: 32px 16px 16px;
+  background: #fff;
+
+  &__images {
+    display: block;
+    margin: 0 auto 40px;
+    width: 160px;
+    height: 134px;
+  }
+
+  &__title {
+    margin: 0 0 4px;
+    font-size: 12px;
+    line-height: 18px;
+    font-weight: 500;
+    color: $text-color;
+    font-family: $main-font-family;
+  }
+
+  &__code {
+    margin-bottom: 16px;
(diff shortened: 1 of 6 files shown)


src/styles/components/card.scss
```diff
@@ -0,0 +1,94 @@
...
+    border: 1px solid $main-color;
+    border-radius: 5px;
+    font-size: 14px;
+    line-height: 16px;
+    color: #fff;
+    font-weight: 700;
+    text-decoration: none;
+    text-transform: uppercase;
+    background-color: $main-color;
+    font-family: $main-font-family;
```
EdPirro, line 87: Instead of setting font-family in every block consider applying it to `body` and use it as the default font for the entire content

src/styles/components/card.scss
```diff
@@ -0,0 +1,94 @@
...
+
+    &:hover {
+      color: $main-color;
+      background-color: #fff;
```
EdPirro, line 91: Hover transitions should be smooth [Reference](

EdPirro (changes requested):
Amazing job, @ropanefarias! Your solution looks very clean and straightforward, I'll request just 2 very minor changes to make your submission simply flawless!

pedro-ruas (approved):
Rodrigo, your code looks great, only thing I'd say you could improve is what was previously highlighted by Eduardo.
You can use smooth hovering effects for better result, I see you added a
`transition: color .3s` to your button at line 83.
You see, **color** in this case applies the transition only to the button text color, not the button background color (you can clearly see this if you change the transition for a longer time, like 2s). The idea is to apply it also to the background color, so you could try using "all" instead

PR #4240, opened 2024-03-11
VB product cards solution

vbuldenko:
- [DEMO LINK](https://vbuldenko.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://vbuldenko.github.io/layout_product-cards/report/html_report/)

❗️ Copy this `Checklist` to the `Pull Request` description after links, and put `- [x]` before each point after you checked it.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -34,9 +34,9 @@
         src="./images/imac.jpeg"
         alt="image of the iMac"
       />
-      <h1 class="card__product-description">
+      <h2 class="card__product-description">
         APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
-      </h1>
+      </h2>
       <p class="card__product-code">Product code: 195434</p>
       <div class="card__reviews">
         <div class="stars stars--4">
--- a/src/index.html
+++ b/src/index.html
@@ -8,11 +8,57 @@
     />
     <title>Product cards</title>
     <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
+      rel="stylesheet"
+    />
+    <link
       rel="stylesheet"
       href="./styles/index.scss"
     />
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div
+      class="card"
+      data-qa="card"
+    >
+      <img
+        class="card__image"
+        src="./images/imac.jpeg"
+        alt="image of the iMac"
+      />
+      <h1 class="card__product-description">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </h1>
+      <p class="card__product-code">Product code: 195434</p>
+      <div class="card__reviews">
+        <div class="stars stars--4">
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+        </div>
+        <p>Reviews: 5</p>
+      </div>
+      <div class="card__price">
+        <p>Price:</p>
+        <p>$2,199</p>
+      </div>
+      <a
+        href="#"
+        data-qa="hover"
+        class="card__button"
+      >
+        Buy
+      </a>
+    </div>
   </body>
 </html>
--- a/src/styles/card.scss
+++ b/src/styles/card.scss
@@ -2,19 +2,17 @@
   display: flex;
   flex-direction: column;
   align-items: center;
-  font-weight: 400;
   width: 200px;
   border: 1px solid rgb(243, 243, 243);
   border-radius: 5px;
+  padding: 32px 16px 16px;
 
   &__image {
-    margin-top: 32px;
     width: 160px;
     height: 134px;
   }
 
   &__product-description {
-    width: 166px;
     height: 36px;
     font-size: 12px;
     line-height: 18px;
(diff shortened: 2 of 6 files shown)


src/styles/card.scss
```diff
@@ -0,0 +1,90 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  font-weight: 400;
+  width: 200px;
+  height: 408px;
```
TarasHoliuk, line 7: do not hardcode height - it should depend on content (and margins between content elements)

src/styles/card.scss
```diff
@@ -0,0 +1,90 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  font-weight: 400;
```
TarasHoliuk, line 5: it's default value - no need to set it

src/styles/card.scss
```diff
@@ -0,0 +1,90 @@
...
+  border-radius: 5px;
+
+  &__image {
+    margin-top: 32px;
+    width: 160px;
+    height: 134px;
+  }
+
+  &__product-description {
+    width: 166px;
```
TarasHoliuk, line 18: it's better to add correct padding to the `.card` and set width `100%` (but in most cases you don't even need to set it since block elements have 100% width by default)

src/styles/card.scss
```diff
@@ -0,0 +1,90 @@
...
+    height: 36px;
+    font-size: 12px;
+    line-height: 18px;
+    font-weight: 500;
+    margin-top: 40px;
+    color: $main-clr;
+  }
+
+  &__product-code {
+    width: 98px;
```
TarasHoliuk, line 28: there is no need to hardcode width

src/styles/card.scss
```diff
@@ -0,0 +1,90 @@
...
+    & :last-child {
+      font-weight: 700;
+      font-size: 16px;
+      color: $main-clr;
+    }
+  }
+
+  &__button {
+    margin-top: 16px;
+    width: 166px;
```
TarasHoliuk, line 71: same about width

TarasHoliuk (changes requested):

src/styles/card.scss
vbuldenko: Cannot get what exactly can have a default value here. Please give broader suggestions.

src/styles/card.scss
TarasHoliuk: still not fixed:
image

If you have a question related to this comment (or any other question) feel free to ask in the chat

src/styles/card.scss
TarasHoliuk: `400` is default value for [font-weight](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#normal)

src/styles/card.scss
```diff
@@ -0,0 +1,88 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  font-weight: 400;
```
TarasHoliuk, line 5:
```suggestion
```

src/styles/card.scss
```diff
@@ -0,0 +1,88 @@
...
+    & :last-child {
+      font-weight: 700;
+      font-size: 16px;
+      color: $main-clr;
+    }
+  }
+
+  &__button {
+    margin: 16px 0;
+    width: 166px;
```
TarasHoliuk, line 69: still not fixed:
image

TarasHoliuk (changes requested):

src/index.html
```diff
@@ -7,12 +7,58 @@
...
+    <div
+      class="card"
+      data-qa="card"
+    >
+      <img
+        class="card__image"
+        src="./images/imac.jpeg"
+        alt="image of the iMac"
+      />
+      <h1 class="card__product-description">
```
Viktor-Kost, line 37: it's better to use `h2` since `h1` we use as a page title
```suggestion
      <h2 class="card__product-description">
```

src/styles/card.scss
```diff
@@ -0,0 +1,86 @@
...
+    justify-content: space-between;
+    align-items: center;
+    font-size: 10px;
+    color: $main-clr;
+    margin-top: 16px;
+  }
+
+  &__price {
+    width: 100%;
+    height: 18px;
```
Viktor-Kost, line 46: no needs to hardcode height. previous mentor explained about it
```suggestion
```

src/styles/card.scss
```diff
@@ -0,0 +1,86 @@
...
+  border-radius: 5px;
+  padding: 32px 16px 16px;
+
+  &__image {
+    width: 160px;
+    height: 134px;
+  }
+
+  &__product-description {
+    height: 36px;
```
Viktor-Kost, line 16:
```suggestion
```

src/styles/index.scss
```diff
@@ -1,3 +1,9 @@
-body {
+@import 'variables';
+@import 'card';
+@import 'stars';
+
+body * {
+  font-family: Roboto, sans-serif;
+  box-sizing: border-box;
```
Viktor-Kost, line 7: it's allowed to move only this rule to `*` selector

src/styles/card.scss
```diff
@@ -0,0 +1,86 @@
...
+    font-size: 10px;
+    line-height: 14px;
+    color: $accent-clr;
+    margin-top: 4px;
+    align-self: flex-start;
+  }
+
+  &__reviews {
+    width: 100%;
+    height: 16px;
```
Viktor-Kost, line 35:
```suggestion
```

Viktor-Kost (changes requested):
Nice!
A few moments need to be fixed. Previous mentor asked you not to hardcode height

src/styles/index.scss
vbuldenko: Did not quite get what was wrong here. If it was about deleting margin values, tests would not pass.

etojeDenys (approved):
lgtm

PR #4250, opened 2024-03-13
Sonic

yurii-filiushkin:
- [DEMO LINK](https://yurii-filiushkin.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://yurii-filiushkin.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,6 +6,19 @@
       name="viewport"
       content="width=device-width, initial-scale=1.0"
     />
+    <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="anonymous"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
+      rel="stylesheet"
+    />
     <title>Product cards</title>
     <link
       rel="stylesheet"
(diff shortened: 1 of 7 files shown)


src/index.html
```diff
@@ -6,13 +6,67 @@
...
+      <div class="card__article">Product code: 195434</div>
+
+      <div class="card__reviews">
+        <div class="stars">
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+        </div>
```
DanilWeda, line 52: Try to use ul and li in this case.
it's preferred because we have one repeated element inside div.
In this case best variant use list)

src/styles/stars.scss
DanilWeda, line 1: All good. 
But for scaling, you can write all cases ( modifiers) for stars. 
In the task, we have only a one starsline ( 4 stars) but in the future, we can add a product card with 1 or 2 stars. 
So, you can inherit your previous task, but create one modifier for 4 stars, and then, if we need 1 or 2 stars, add new selectors in this part of the code)
Hope you correctly understand what I mean. If you have additional questions about it, write me a comment) 
Many thanks!

DanilWeda (changes requested):
Good job! 
Check some small improvements. 
Many thanks)

DanilWeda (changes requested):
rc*

src/styles/stars.scss
yurii-filiushkin: Thanks for advices. With loop and ul it is much better.

DanilWeda (approved):
Great! 
Approved!

PR #4285, opened 2024-03-21
Develop

RafaelPCouto:
- [DEMO LINK](https://RafaelPCouto.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://RafaelPCouto.github.io/layout_product-cards/report/html_report/)

- [ ] there 2 BEM blocks card and stars each in it own file
- [ ] SCSS Nesting is used for elements, modifiers and pseudo-classes
- [ ] SCSS Variables are used for main values and placed in a separate file
- [ ] all stars--N modifiers work as expected (Highlight first N stars)
- [ ] Code follows all the [Code Style Rules ❗️](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules)

### DIFF
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -1,11 +1,9 @@
-$font: 'Roboto', sans-serif;
-$primaryColor: #060b35;
-$secondaryColor: #616070;
-$blueAccent: #00acdc;
-$whiteAccent: #fff;
+@import 'variables';
+@import 'star';
 
 body {
   margin: 0;
+  padding: 0;
 }
 
 .card {
@@ -15,19 +13,32 @@ body {
   box-sizing: border-box;
   display: flex;
   flex-direction: column;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
 
-  &__image {
+  &__img-container {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+  }
+
+  &__img {
     width: 160px;
     height: 134px;
-    padding-inline: 3px;
+  }
+
+  &__infos {
+    margin-top: 40px;
   }
 
   &__name {
     font-family: $font;
+    color: $primaryColor;
     font-weight: 500;
     font-size: 12px;
     line-height: 18px;
-    color: $primaryColor;
+    text-align: left;
+    margin: 0;
   }
 
   &__code {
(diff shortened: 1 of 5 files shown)


src/styles/index.scss
```diff
@@ -1,3 +1,108 @@
...
+@import 'star';
+
 body {
   margin: 0;
+  padding: 0;
+}
+
+.card {
+  width: 200px;
+  height: 408px;
```
anastasiia-tilikina, line 11: do not hardcode height for card, let content define it

src/styles/index.scss
```diff
@@ -1,3 +1,108 @@
...
+  &__button {
+    background-color: $blueAccent;
+    color: $whiteAccent;
+    border: none;
+    border-radius: 5px;
+    text-align: center;
+    text-decoration: none;
+    box-sizing: border-box;
+    width: 166px;
+    height: 40px;
```
anastasiia-tilikina, line 98: don't hardcode height here as well
use proper padding instead
```suggestion
    width: 100%;
```

src/styles/star.scss
```diff
@@ -0,0 +1,16 @@
...
+.stars__star {
+  background-image: url(../images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+}
+
+.stars__star:last-child {
```
anastasiia-tilikina, line 14: You should move all modifiers from the previous task

src/styles/index.scss
```diff
@@ -1,3 +1,108 @@
+@import 'variables';
+@import 'star';
+
 body {
   margin: 0;
+  padding: 0;
+}
+
+.card {
```
anastasiia-tilikina, line 9: it is better to create separate file for each BEM block and name it respectively
in your index.scss file you can keep imports and some common styles like for body or HTML tag or resetting some default styles

anastasiia-tilikina (changes requested):
Good job overall!
Here are some suggestions:

It is better to reset default browser styles like margins and define spacing on your own
image

in your case you can reset margin for all p tags

Your button font-family is wrong
image

lerastarynets (approved):
You're on the right track!

lerastarynets (approved):
Keep up the great work!

PR #4294, opened 2024-03-25
Product cards

Teslaanazuma:
Requirements:
reset browser's default margins
card width is 200px including border
use images from [src/images](https://github.com/Teslaanazuma/layout_product-cards/blob/master/src/images)
change link styles on :hover
add data-qa="card" attribute to the card block
add data-qa="hover" attribute to the link BUY
background-image: url() should be relative to the index.scss. So should start with ../images.
Rewrite stars block from the [Stars task](https://github.com/mate-academy/layout_stars) with SCSS and use it
Find the required font on [google fonts](https://fonts.google.com/) and use.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "@mate-academy/backstop-config": "latest",
     "@mate-academy/bemlint": "latest",
     "@mate-academy/linthtml-config": "latest",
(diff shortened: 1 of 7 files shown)


etojeDenys (approved):
lgtm

PR #4324, opened 2024-04-03
Added Solution

FurArt:
Hello,
- [DEMO LINK](https://furart.github.io/layout_product-cards/)


src/styles/main.scss
```diff
@@ -0,0 +1,103 @@
...
+}
+
+.product {
+  @include flex-column;
+
+  width: $product-full-width;
+  background-color: $background;
+  align-items: center;
+  text-align: center;
+  padding-block: 34px 0;
```
vadiimvooo, line 15: you have to add paddings according to the mockup. Paddings should be used as part of card and not as margins like in your solution right now.

src/styles/main.scss
```diff
@@ -0,0 +1,103 @@
...
+  &-image {
+    width: $product-full-width * 0.8;
```
vadiimvooo, line 17: This value should have fixed width in pixels without any calculations. Please check the mockup.

vadiimvooo: The same for other elements.

src/styles/main.scss
```diff
@@ -0,0 +1,103 @@
...
+    color: #060b35;
+    font-weight: 400;
+  }
+
+  &-price {
+    display: flex;
+    justify-content: space-between;
+
+    &-text {
+      font-family: Roboto, sans-serif;
```
vadiimvooo, line 64: You don't need to doublicate font-family property. Just use it in for parent (element with class card)

vadiimvooo: The same is for other places where you using font-family parametr.

vadiimvooo (changes requested):
Good job. But still you need to fix the tests. There are some fixes need to be done for it. Please feel free to ask for help in fe_chat or QnA session if you have some question related to the task.

src/styles/reset.scss
```diff
@@ -0,0 +1,86 @@
...
+h3,
+h4,
+h5,
+h6,
+p,
+blockquote,
+pre,
+a,
+abbr,
+acronym,
```
volodymyr-soltys97, line 19: Do you so many tag for reset?
Use reset only for those tags that you use on the page
```suggestion
html,
body,
div,
span,
applet,
object,
iframe,
h1,
h2,
h3,
h4,
h5,
h6,
p,
blockquote,
pre,
a,
abbr,
acronym,
```

src/styles/variables.scss
```diff
@@ -0,0 +1,16 @@
+$product-full-width: 200px;
+$content-width: 166px;
```
volodymyr-soltys97, line 2:
```suggestion
$content-width: 100%;
```

volodymyr-soltys97 (changes requested):
GJ!
You need to add to left and right padding for the card and remove it everywhere
image

volodymyr-soltys97 (approved):
Good job 👍

PR #4325, opened 2024-04-03
attempt_1

igols:
https://igols.github.io/layout_product-cards/
https://igols.github.io/layout_product-cards/report/html_report/


vadiimvooo (changes requested):
Good job. Please redeploy your solution cause it seems that styles are not loading. Please feel free to ask for help in fe_chat if you need so.

etojeDenys (changes requested):
try to fix the tests, don't hesitate to ask for some help about that in the chat

PR #4409, opened 2024-05-03
add task solution

V-Shut:


src/index.html
```diff
@@ -11,8 +11,61 @@
...
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="anonymous"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
```
dinaminahi, line 24: Do we need to import all weights if we use only 3?
```suggestion
     href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
```

src/styles/index.scss
```diff
@@ -1,3 +1,21 @@
+$card-color: #fff;
+$border-color: #f3f3f3;
+$font-main-color: #060b35;
+$font-secondary-color: #616070;
+$inactive-button-color: #00acdc;
+$active-button-color: #fff;
+$inactive-button-font-color: #fff;
+$active-button-font-color: #00acdc;
+$active-button-border-color: #00acdc;
```
dinaminahi, line 9: And let's create a separate file for these variables and import it here

dinaminahi, line 9: there is no sense to create two variables with different names for one variable
The sense of variables is to have a value in one place so that you can change it in one place and it will change everywhere

What if you used #fff 100 times through your project,  would you create 100 different names for it? or just one and use it everywhere?

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,86 @@
...
+    color: $font-main-color;
+    height: 36px;
+    width: 166px;
+    margin-top: 40px;
+    margin-bottom: 0;
+  }
+
+  &__description {
+    font-size: 10px;
+    font-weight: 400;
```
dinaminahi, line 31: This is default value for font-weight, so we don't have to write it
```suggestion
```

dinaminahi (changes requested):
Good job 👍 
But we need to make some improvements

src/styles/index.scss
V-Shut: I already tried to create a separate file, but after that “npm start” command can’t start browser page, just infinite loading. “npm test” can’t start too…

src/styles/blocks/card.scss
V-Shut: Must I convert scss to css?

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  &__reviews {
+    margin-top: 16px;
+    margin-bottom: 24px;
+    height: 16px;
+    width: 166px;
+
+    font-size: 10px;
+    font-weight: 400;
```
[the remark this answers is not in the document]
V-Shut: Ok) Thanks)

PR #4413, opened 2024-05-05
product card

LitviVit:
- [DEMO LINK](https://LitviVit.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://LitviVit.github.io/layout_product-cards/report/html_report/)

* box-sizing: border-box [^1].

[^1]: the property does not greatly affect performance. In the opinion of the coach, it is allowed for use in the selector *

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,17 +2,49 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
-    <meta
-      name="viewport"
-      content="width=device-width, initial-scale=1.0"
-    />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Product cards</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" />
     <link
       rel="stylesheet"
-      href="./styles/index.scss"
+      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
     />
+    <link rel="stylesheet" href="./styles/index.scss" />
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div class="card" data-qa="card">
+      <img
+        src="./images/imac.jpeg"
+        alt="APPLE A1419 iMac 27' Retina 5K Monoblock"
+        class="card__image"
+      />
+
+      <h1 class="card__title">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </h1>
+
+      <p class="card__description">Product code: 195434</p>
+
+      <div class="card__review">
+        <div class="stars stars--4">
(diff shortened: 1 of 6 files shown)


LitviVit:
image

src/index.html
```diff
@@ -2,17 +2,49 @@
...
-      content="width=device-width, initial-scale=1.0"
-    />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Product cards</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" />
     <link
       rel="stylesheet"
-      href="./styles/index.scss"
+      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
```
roman-mirzoian, line 11: Do we need to import all the weights if we are only using 3?
```suggestion
      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
```

src/styles/vars.scss
```diff
@@ -0,0 +1,7 @@
+$accent_color: #00acdc;
+$gray_elements: #f3f3f3;
+$primary-color: #060b35;
+$secondary_color: #616070;
```
roman-mirzoian, line 4: It is also better to set all naming in the same style.
```suggestion
$gray_elements: #f3f3f3;
$primary_color: #060b35;
$secondary_color: #616070;
```

src/styles/vars.scss
```diff
@@ -0,0 +1,7 @@
...
+$white_color: #fff;
+$title_size: 12px;
+$desc_size: 10px;
```
roman-mirzoian, line 7: It is better to use more generic variable names. 

```suggestion
$accent_color: #00acdc;
$second_color: #f3f3f3;
$primary_color: #060b35;
$secondary_color: #616070;
$button_color_active: #fff;
$title_size: 12px;
$description_size: 10px;
```

src/styles/index.scss
```diff
@@ -1,3 +1,15 @@
+* {
```
roman-mirzoian, line 1: Don't use * selector (it impacts performance), better use ``html`` or ``body`` tag.

image

roman-mirzoian (changes requested):
Well done, but pay attention to this comments.

src/styles/index.scss
roman-mirzoian: @LitviVit I read your comment carefully and that's why I added a screenshot of the checklist: using the generic selector "*" is not a good practice, so it's best to avoid using this approach.
Each mentor has personal experience depending on the company where he works, so I focus on the general checklist during the inspection so that there are fewer questions.

LitviVit: @roman-mirzoian ok. Understood. So what do you think about this solution?
image
I found it while researching this question on stackoverflow

roman-mirzoian: @LitviVit This solution is better than specifying this rule directly 👍

PR #4428, opened 2024-05-10
Develop

MarinaTop:
- [DEMO LINK](https://MarinaTop.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://MarinaTop.github.io/layout_product-cards/report/html_report/)

- [x] there 2 BEM blocks card and stars each in it own file
- [x] SCSS Nesting is used for elements, modifiers and pseudo-classes
- [x] SCSS Variables are used for main values and placed in a separate file
- [x] all stars--N modifiers work as expected (Highlight first N stars)

### DIFF
--- a/src/styles/stars.scss
+++ b/src/styles/stars.scss
@@ -1,0 +1,27 @@
+.stars {
+  width: 96px;
+  height: auto;
+
+  display: flex;
+  flex-direction: row;
+  justify-content: space-around;
+
+  &__star {
+    width: 16px;
+    height: 16px;
+    background-image: url(../images/star.svg);
+    background-position: left;
+    background-repeat: no-repeat;
+    background-size: 80%;
+
+    &:not(:last-child) {
+      margin-right: 4px;
+    }
+  }
+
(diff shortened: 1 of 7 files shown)


src/styles/stars.scss
```diff
@@ -0,0 +1,27 @@
...
+
+    &:not(:last-child) {
+      margin-right: 4px;
+    }
+  }
+
+  &--4 {
+    .stars__star:nth-child(-n + 4) {
+      background-image: url(../images/star-active.svg);
+    }
```
DarkMistyRoom, line 25: Use the whole `stars` block from the previous task. It should be possible to set any amount of stars

DarkMistyRoom (changes requested):
Good job! But here are some comments:
- Be consistent with your margins. Add only top or bottom, don't add both.
image

- Would be better to use vertical padding for the parent element as well as horizontal instead of adding margins to two side elements
image

nazarmatsevych (approved):
Well done 🔥

PR #4461, opened 2024-05-28
Add solution

ivanKo95:


src/styles/index.scss
```diff
@@ -1,3 +1,102 @@
...
+
+.card {
+  width: 200px;
+  padding: 32px 16px 0;
+  border-radius: 5px;
+  align-items: center;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  opacity: 1;
```
volodymyr-soltys97, line 28: Redundant
```suggestion
```

src/styles/index.scss
```diff
@@ -1,3 +1,102 @@
...
+      line-height: 18px;
+      color: $textColor;
+    }
+  }
+
+  &__button {
+    margin-bottom: 16px;
+    background: $blueColor;
+    color: #fff;
+    width: 166px;
```
volodymyr-soltys97, line 90:
```suggestion
    width: 100%;
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
This part of the task is incomplete, you need to use stars styles from the previous task 
image
image

src/styles/index.scss
ivanKo95: I've deleted this style.

src/styles/index.scss
ivanKo95: I've updated it to 100%.

volodymyr-soltys97 (changes requested):
Hi, this comment still not fixed from the previous review, if you need help feel free in the fe_chat 
image

lerastarynets (approved):
Keep up the great work!

PR #4486, opened 2024-06-08
add solution

RimmaBuzuluk:
- [DEMO LINK](https://RimmaBuzuluk.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://RimmaBuzuluk.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -1,3 +1,140 @@
 body {
   margin: 0;
+  font-family: Roboto, sans-serif;
 }
+
+ul {
+  list-style-type: none;
+  padding: 0;
+  margin: 0;
+}
+
+ul li {
+  padding: 0;
+  margin: 0;
+}
+
+.card {
+  width: 200px;
+  height: 408px;
+  border: 1px solid #f3f3f3;
+  background-color: #fff;
+  border-radius: 5px;
+  box-sizing: border-box;
+  padding: 32px 16px 16px;
+
+  // &__img {
+  //   display: block;
+  //   width: 160px;
+  //   height: 134px;
+  //   background-image: url(../images/imac.jpeg);
+  //   background-position: center;
+  //   background-repeat: no-repeat;
+  //   background-size: contain;
+  //   bac
+  //   margin: 0 3px 40px;
+
+  //   // margin-bottom: 40px;
+  // }
+  &__img {
+    margin: 0 2px 0 3px;
+    img {
+      height: 134px;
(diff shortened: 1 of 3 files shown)


src/styles/index.scss
```diff
@@ -1,3 +1,140 @@
...
+
+    &:hover {
+      background-color: #fff;
+      border: 1px solid #00acdc;
+      color: #00acdc;
+    }
+  }
+}
+
+.stars {
```
DanilWeda, line 116: You need realize all variants of stars (1,2,3..). Please check scss loop and try to do it.

src/styles/index.scss
```diff
@@ -1,3 +1,140 @@
...
+    img {
+      height: 134px;
+      width: 160px;
+
+      margin-bottom: 36px;
+    }
+  }
+
+  &__title {
+    color: #060b35;
```
DanilWeda, line 50: Also, you can create separate file with styles, and move all colors in variables, then import and use in all places of your project.

src/styles/index.scss
```diff
@@ -1,3 +1,140 @@
...
+  margin: 0;
+}
+
+ul li {
+  padding: 0;
+  margin: 0;
+}
+
+.card {
+  width: 200px;
```
DanilWeda, line 18: Try not use static width/height in places where their is not critical need to use.

DanilWeda (changes requested):
Good! 
Thanks for your request! 
I left some small comments, for improving your solution. 
Please, check it.
Many thanks)

src/styles/index.scss
```diff
@@ -1,3 +1,137 @@
...
+    height: 16px;
+    background-repeat: no-repeat;
+    background-position: center;
+
+    &:not(:last-child) {
+      margin-right: 4px;
+    }
+  }
+
+  &--1 .stars__item:nth-child(-n + 1) {
```
DanilWeda, line 118: All good but in this case you can use loop construction 
https://sass-lang.com/documentation/at-rules/control/for/
Which made you index for prefixes and nth-child.
@for $i from 1 through 5 {
 then use $i which will grow from 1 to 5.
}

DanilWeda (changes requested):
All good! Great! 
Check a minor comment about using scss loop. 
Thanks)

DanilWeda (approved):
Great. 
Approved! 
Thanks)

PR #4519, opened 2024-06-19
all requirements are performed

NIKaragu:
- [DEMO LINK](https://NIKaragu.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://NIKaragu.github.io/layout_product-cards/report/html_report/)
=================================================================================
=================================================================================
***A question from me:*** my code seems to me a little awkward (especially card.scss), **how can I improve it and do it a little more elegant**?

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,90 @@
+.card__wrapper {
+  margin: 0 16px;
+}
+
+.card {
+  display: flex;
+  flex-direction: column;
+  width: 200px;
+  height: 408px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  box-sizing: border-box;
+
+  &__image {
+    width: 160px;
+    height: 134px;
+    background-image: url('/src/images/imac.jpeg');
+    background-size: 160px 134px;
+    margin: 32px 19px 40px;
+  }
+
+  &__name {
+    font-weight: 500;
+    font-size: 12px;
+    line-height: 18px;
+    margin-bottom: 4px;
+    color: $info-product-color;
+  }
+
+  &__code {
+    color: $identificator-color;
+    line-height: 14px;
+    margin-bottom: 16px;
+  }
+
+  &__rating {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    margin-bottom: 24px;
+  }
+
+  &__reviews {
+    color: $info-product-color;
+  }
+
+  &__price {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    line-height: 18px;
+    margin-bottom: 16px;
+
+    &-text {
+      font-size: 12px;
+      color: $identificator-color;
+    }
+
+    &-price {
+      font-size: 16px;
+      font-weight: 700;
+      color: $info-product-color;
+    }
+  }
+
+  &__button {
+    background-color: #00acdc;
+    font-size: 14px;
+    height: 40px;
+    width: 166px;
+    border: 1px solid #00acdc;
+    border-radius: 5px;
+  }
+
+  &__button:hover {
+    transition: 0.12s;
+    background-color: white;
+    color: #00acdc;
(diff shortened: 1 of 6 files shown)


src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  &__star:not(:last-child) {
+    margin-right: 4px;
+  }
+
+  &--4 .stars__star:nth-child(-n + 4) {
+    background-image: url(/src/images/star-active.svg);
+    background-repeat: no-repeat;
+  }
```
ArturErmolenko, line 19: all stars--N modifiers work as expected (Highlight first N stars)
 
 What if I change the modifier to 3 will it work?)

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,91 @@
...
+  &__button {
+    background-color: #00acdc;
+    font-size: 14px;
+    height: 40px;
+    width: 166px;
+    border: 1px solid #00acdc;
+    border-radius: 5px;
+
+    &:hover {
+      transition: 0.12s;
```
ArturErmolenko, line 75: the transition of what?) you should directly identify the property that animated

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,91 @@
+.card__wrapper {
+  margin: 0 16px;
+}
+
+.card {
+  display: flex;
+  flex-direction: column;
+  width: 200px;
+  height: 408px;
```
ArturErmolenko, line 9: Don't fix container size (if there is no such a requirement)
Let the content size dictate it. To avoid overflow or accidental scroll bar

ArturErmolenko, line 9:
```suggestion
```

ArturErmolenko (changes requested):
Well done 🔥 
Some small improvements 🐝

src/styles/blocks/stars.scss
NIKaragu: Good notice) I'll changed it with SASS

src/styles/blocks/card.scss
NIKaragu: Okey, I'll correct this moment

src/styles/blocks/stars.scss
NIKaragu: UPD: I've resolved that problem with scss, but then I realized I also have to change index.html accordingly to changes, but I can't make that without JS. It's so far to this JS topic, unforrtunately.
But I will be glad if you can give me a tip :D

DanilWeda (changes requested):
Good job. 

Please check this UI moment (hover on button) : 
image

Also, in transition you need use property ( for example color, opacity). 
Default value for transition is all, but we don't recommended use it, because this  approach has bad performance and has side effects.

If you have additional question, you can write on platform. 
Many thanks)

NIKaragu:
> Good job.
> 
> Please check this UI moment (hover on button) : image
> 
> Also, in transition you need use property ( for example color, opacity). Default value for transition is all, but we don't recommended use it, because this approach has bad performance and has side effects.
> 
> If you have additional question, you can write on platform. Many thanks)

I don't know why it doesn't work in your case, but for me everything is good
image
image

ArturErmolenko:
> > Good job.
> > Please check this UI moment (hover on button) : image
> > Also, in transition you need use property ( for example color, opacity). Default value for transition is all, but we don't recommended use it, because this approach has bad performance and has side effects.
> > If you have additional question, you can write on platform. Many thanks)
> 
> I don't know why it doesn't work in your case, but for me everything is good image image
> 
> Your recommendation I got and I corrected the moment with transition, thanks :)

Probably you forgot to redeploy your code, because Demo link still show incorrect behaviour

ArturErmolenko (approved):
GJ 🔥 
I left few comments what could be improved, but you can fix it in next PR with cards catalog

PR #4526, opened 2024-06-20
Develop

totum-revolutum:
Hello! I was unable to fix the tests to pixel-perfect. Аt the Q&A session solved this issue with a mentor and came to the conclusion that we need to leave it like that.

- [DEMO LINK](https://totum-revolutum.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://totum-revolutum.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,8 +11,53 @@
       rel="stylesheet"
       href="./styles/index.scss"
     />
+    <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,400;0,500;0,700;1,400;1,500;1,700&display=swap"
+      rel="stylesheet"
+    />
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div
+      data-qa="card"
+      class="card"
+    >
+      <img
+        class="card__img"
+        src="./images/imac.jpeg"
+        alt="imac image"
+      />
+      <div class="card__name">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </div>
+      <div class="card__code">Product code: 195434</div>
+      <div class="card__raiting">
+        <ul class="stars">
+          <li class="stars__star"></li>
+          <li class="stars__star"></li>
+          <li class="stars__star"></li>
+          <li class="stars__star"></li>
+          <li class="stars__star"></li>
+        </ul>
+        <div>Reviews: 5</div>
+      </div>
+      <div class="card__price">
+        Price:
+        <div class="card__price_value">$2,199</div>
+      </div>
+      <button
+        data-qa="hover"
+        class="card__btn"
+      >
+        Buy
+      </button>
+    </div>
   </body>
 </html>
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,76 @@
+.card {
+  width: 200px;
+  height: 406px;
+  display: flex;
+  flex-direction: column;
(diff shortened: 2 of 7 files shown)


src/index.html
```diff
@@ -11,8 +11,53 @@
...
+    >
+      <img
+        class="card__img"
+        src="./images/imac.jpeg"
+        alt="imac image"
+      />
+      <div class="card__name">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </div>
+      <div class="card__code">Product code: 195434</div>
```
BudnikOleksii, line 40: - Use semantic tags where possible
image

- alt attribute should describe the image content
image

src/index.html
```diff
@@ -11,8 +11,53 @@
...
+      <div class="card__raiting">
+        <ul class="stars">
```
BudnikOleksii, line 42:
```suggestion
        <ul class="stars stars--4">
```
You should have some modifier to mark active stars inside depending on it

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,76 @@
+.card {
+  width: 200px;
+  height: 406px;
```
BudnikOleksii, line 3: - Don't fix container size (if there is no such a requirement) - in this case don't hardcode height
image

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,76 @@
...
+    color: white;
+    height: 40px;
+    border-radius: 5px;
+    border: none;
+    font-size: 14px;
+    font-weight: 700;
+    text-transform: uppercase;
+    padding: 0;
+    margin-top: 16px;
+    font-family: Roboto, sans-serif;
```
BudnikOleksii, line 68: It's your main font that you have specified to body, so you don't need to set it here

BudnikOleksii (changes requested):
Good work 👍 
But could be even better, check the comments and improve your solution. Also, don't forget about the checklist and copy it to the PR description(and mark what you have done from it).
image

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,21 @@
...
+    background-repeat: no-repeat;
+    background-position: center;
+    height: 16px;
+    width: 16px;
+    margin-right: 4px;
+  }
+
+  &__star:not(:last-child) {
+    background-image: url(../images/star-active.svg);
+  }
```
ArturErmolenko, line 20: if I change modifier nothing happens(
Reuse styles from Stars task so that modifier functionality work as expected

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,4 @@
+$main-text-color: #060b35;
+$sec-text-color: #616070;
+$btn-color: #00acdc;
```
ArturErmolenko, line 3: better use more descriptive name for colors for example $color-blue 
Because this color can be used in another place not only button)

ArturErmolenko (changes requested):
Almost done 🐝 
Let's fix the stars functionality(active stars should change if modifier was changed)

https://github.com/mate-academy/layout_product-cards/assets/94292654/648d1051-3986-4780-90be-89d29a443975

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,25 @@
...
+    width: 16px;
+    margin-right: 4px;
+  }
+}
+
+.stars--1 :first-child,
+.stars--2 :nth-child(-n + 2),
+.stars--3 :nth-child(-n + 3),
+.stars--4 :nth-child(-n + 4),
+.stars--5 :nth-child(-n + 5) {
```
DanilWeda, line 23: You can use scss loop in this case ( for future)

DanilWeda (approved):
Good job. Approved. 
Also, you can add a cursor pointer for the button ( because it an action element) 
Thanks)

PR #4586, opened 2024-07-12
add task solution

Matheusgroff:
- [DEMO LINK](https://Matheusgroff.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://Matheusgroff.github.io/layout_product-cards/report/html_report/)

I had problems with the stars box, the mixins for card and stars and once again Backstop didn't help me.


gitlherme (changes requested):
Tip: Don't always take Figma as the "source of truth", always rely on the Backstop test as well to be able to pass the tests correctly.

- The background color of the button should be: #00ACDC;
- The font should be Roboto, you need to import it into your CSS for it to work, get it from Google Fonts. Also check the weight of the font as it doesn't seem to match the image you're testing.
- Also check the spacing between your elements, they look a bit different from the test image.
- You are rewriting code and perhaps this is causing a conflict between your styles. To create the card and stars, create separate files and import them into your `index.scss` (https://sass-lang.com/documentation/at-rules/import/). Remember to put _ at the beginning of the file to signal that it will be imported (`_card.scss` and `_stars.css` for example).

gitlherme (changes requested):
@Matheusgroff  Looking at your tests, you're almost there. If you see where I've marked it in yellow, it says that the y-axis, i.e. vertically, has a difference of 2 between the test and the reference. In other words, between what you did and what they use as a basis for the test.
In the second image, I've left the reference on the left and the test, which is yours, on the right. You can see that when I crossed the straight line, your price is a little lower than the reference. I think that's what's causing the test not to pass. It really is a small difference, but the backstop only accepts a 1% difference between the reference and the test. And yours is reaching 5%. I think you can finish the exercise if you fix this, but if you need anything else, let me know.

image

image

PR #4616, opened 2024-07-28
upload tarefa product-cards

14lucas-mendes:
[DEMO LINK](https://14lucas-mendes.github.io/layout_product-cards/)
[TEST REPORT LINK](https://14lucas-mendes.github.io/layout_product-cards/report/html_report/)


src/styles/index.scss
```diff
@@ -1,3 +1,145 @@
...
+  border-radius: 5px;
+  margin-left: 16px;
+  margin-bottom: 16px;
+}
+
+a {
+  text-decoration: none;
+}
+
+.text--buttom {
```
gitlherme, line 128: Make sure that you are using BEM correctly. The -- is used to modifiers and it is not the case.

Try to use button__text instead of text--buttom.

package.json
gitlherme, line 1: Seems like the scripts and dependencies were removed. I think that you need to keep all the scripts and dependencies. It is not recommended to remove things on package.json.

gitlherme (changes requested):
Hello, @14lucas-mendes!

Just a point about the code. Can you please check it?

But I saw that you changed a lot your package.json. Please go back to the last version. It is not recommended to change package.json.

Probably, this is what is making your deploy not work well.

package.json
pedro-ruas, line 1: Lucas, did you replace this project's file with the file "package.json" from another project?
Seems you added and removed a lot of dependencies and changed the scripts, this was not necessary for this project.
I suggest using the original file, as this may cause several issues.

src/index.html
```diff
@@ -13,6 +13,40 @@
     />
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div
+      class="card card--image"
+      data-qa="card"
+    >
+      <div class="card__text">
+        <p class="card__text1">
```
pedro-ruas, line 21: "text1" is not a great naming. You could probably cal it something like `card__title` or `card__product-name`.
Avoid using multiple names that differ only by a number, for example "text1" and "text2", those really are not the best approach as they indicate both texts have the same purpose, and it's not the case here.

src/index.html
```diff
@@ -13,6 +13,40 @@
...
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </p>
+        <p class="card__text2">Product code: 195434</p>
+        <div class="content">
+          <ul class="content__stars">
+            <li class="stars"></li>
```
pedro-ruas, line 27: "stars" is not a _Block_, right? If it is an _Element_ of "content" the correct naming is `content__star`
Also, changing from stars to star, without the S, as the class refers to a single star and not to all of them

src/index.html
```diff
@@ -13,6 +13,40 @@
...
+    <div
+      class="card card--image"
+      data-qa="card"
+    >
+      <div class="card__text">
+        <p class="card__text1">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </p>
+        <p class="card__text2">Product code: 195434</p>
+        <div class="content">
```
pedro-ruas, line 25: If "content" is a _Block_ that is child of another _Block_ called "card", the name here should also contain the parent, like so: `card__content content`

src/styles/index.scss
```diff
@@ -1,3 +1,145 @@
...
+  background: #fff;
+  width: 198px;
+  height: 406px;
+  top: 1px;
+  left: 1px;
+  border-radius: 5px;
+  border: 1px solid #f3f3f3;
+}
+
+.card--image {
```
pedro-ruas, line 22: One of the greatness of using SCSS instead of CSS, i the fact that you can use a concept called _Nesting_.
Nesting means that you can join class names to reduce code, and you could do it here. Foe example, for all _Elements_ that you have for a _Block_ you can use an "&" to join it.
Example:
```
.card {
  ...<card properties>
  
  &__image {
    ...<card__image properties>
  }

  &__text {
    ...<card__text properties>
  }
}
```
You could try doing so to all your classes that can, if you want

src/styles/index.scss
```diff
@@ -1,3 +1,145 @@
...
+
+.button__text {
+  color: #fff;
+  font-size: 14px;
+  font-weight: 700;
+  text-align: center;
+  padding-top: 12px;
+}
+
+.buttom:hover {
```
pedro-ruas, line 136: You can also use _Nesting_ for `&:hover`, so:

`.button:hover`
is the same as
```
.button {
  &:hover {
 }
}
```

src/styles/index.scss
```diff
@@ -1,3 +1,145 @@
...
+}
+
+.stars {
+  background-image: url(../images/star-active.svg);
+  background-repeat: no-repeat;
+  width: 16px;
+  height: 16px;
+}
+
+.content__stars .stars:last-child {
```
pedro-ruas, line 72: You can also use a class inside another in nesting for this:
```
.content__stars {
  .stars:last-child {
  }
}
```

pedro-ruas (changes requested):
Some changes are necessary for BEM naming to be correct, and nesting can be applied to SCSS

PR #4634, opened 2024-08-04
Develop

nex1994:
- [DEMO LINK](https://nex1994.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://nex1994.github.io/layout_product-cards/report/html_report/)


src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,39 @@
...
+    background-image: url(/src/images/star.svg);
+  }
+
+  &--1 &__star:nth-child(-n + 1),
+  &--2 &__star:nth-child(-n + 2),
+  &--3 &__star:nth-child(-n + 3),
+  &--4 &__star:nth-child(-n + 4),
+  &--5 &__star:nth-child(-n + 5) {
+    background-image: url(/src/images/star-active.svg);
+  }
```
BogdanMaliuta, line 31: dont know for what you need this, all starts should be active and only last not, so change it

BogdanMaliuta (changes requested):
Looks good, but cant uderstand styles for starts, just use last-child

src/styles/blocks/stars.scss
nex1994: Hi. One of the requirements of this task is that all of stars--N modifiers should work as expected so that's why all modifiers are there. Should I delete them?

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,32 @@
...
+    margin-right: 4px;
+    background-size: 13px;
+    background-position: center;
+    background-repeat: no-repeat;
+  }
+
+
+  &--4 &__star:nth-last-child(-n + 2) {
+    background-image: url(/src/images/star.svg);
+  }
```
natalia-klonowska, line 20: delete it and move this style (background-image) inside stars__star because nth:child will overwrite it with active stars

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,32 @@
...
+
+  &--4 &__star:nth-child(-n + 4) {
+    background-image: url(/src/images/star-active.svg);
+  }
```
natalia-klonowska, line 24: return selectors for all --N modifiers

natalia-klonowska (changes requested):
add padding for card instead of having right and left margins for each element

natalia-klonowska (approved):
Great work 🚀

natalia-klonowska (changes requested):
fix demo link
image

PR #4664, opened 2024-08-14
finalização do card

michelletani:
- [DEMO LINK](https://michelletani.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://michelletani.github.io/layout_product-cards/report/html_report/)


src/index.html
```diff
@@ -9,10 +9,63 @@
...
+        alt="Imagem IMAC"
+      />
+      <div class="card_content">
+        <h1 class="card_title">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </h1>
+        <p class="card_product_code">Product code: 195434</p>
+
+        <div class="card_reviews">
+          <div class="card_stars">
```
pedro-ruas, line 45: Tani, the ideia is to use the code from the task Stars here, just copying as it is, so it can be easily changed, as it worked there.
In this way, you should rename this block to `card__stars stars stars--4`, that way you are saying that _Stars_ is a -Block_ that is also an _Element_ of the _Block_ _Card_, and with the _Modifier_ "4" to represent 4 stars.
And them, all card_stars__star could be just renamed to `stars__star`

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,16 @@
...
+
+  &__star {
+    width: 16px;
+    height: 16px;
+    margin-right: 4px;
+    background-image: url(../images/star.svg);
+    background-repeat: no-repeat;
+    background-position: center;
+
+    &:nth-child(-n + 4) {
```
pedro-ruas, line 12: In this manner, you are fixing the styling for only 4 stars, you should use the code from the previous task _Stars_ so it can be changed as needed.
This comment just complements the previous one I made.

pedro-ruas (changes requested):
Tani, your code is in fact well written, but there are 2 main issues:
- You missed the usage of the previous Stars block, as requested in the Checklist
- Your links for the DEMO and TEST are not working, I believe you did not run the command `npm run deploy` to generate the links. So I cannot properly test your task as it should.
Please fix those points and it will be perfect. You are doing a great job so far! =D

src/index.html
michelletani: Ok, thank you for the explanation!!! 😄

pedro-ruas (approved):
Nice job, Tani!

PR #4690, opened 2024-08-15
Product cards

AnastasiiaSavrasna:
[DEMO LINK](https://AnastasiiaSavrasna.github.io/layout_product-cards/)
[TEST REPORT LINK](https://AnastasiiaSavrasna.github.io/layout_product-cards/report/html_report/)

[x]  there 2 BEM blocks card and stars each in it own file
[x]  SCSS Nesting is used for elements, modifiers and pseudo-classes
[x]  SCSS Variables are used for main values and placed in a separate file
[x]  all stars--N modifiers work as expected (Highlight first N stars)
[x]  Code follows all the [Code Style Rules ❗️](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules)

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,93 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  width: 20rem;
+  padding: 3.2rem 1.6rem 1.6rem;
+  border: 0.1rem solid $card-border-color;
+  border-radius: 0.5rem;
+  background-color: $main-background-color;
+
+  &__img {
+    width: 16rem;
+    height: 13.4rem;
+    margin-bottom: 4rem;
+  }
+
+  &__title {
+    font-weight: 500;
+    font-size: 1.2rem;
+    line-height: 1.5;
+    margin-bottom: 0.4rem;
+  }
+
+  &__product-code {
+    align-self: flex-start;
+    font-weight: 400;
+    font-size: 1rem;
+    line-height: 1.4;
+    margin-bottom: 1.6rem;
+    color: $secondary-text-color;
+  }
+
+  &__review-section {
+    display: flex;
+    width: 100%;
+    justify-content: space-between;
+    align-items: center;
+    margin-bottom: 2.4rem;
+  }
+
+  &__reviews {
+    font-weight: 400;
+    font-size: 1rem;
+    line-height: 1.4;
+    align-self: flex-end;
+  }
+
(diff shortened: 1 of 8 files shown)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,93 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  width: 20rem;
+  padding: 3.2rem 1.6rem 1.6rem;
+  border: 0.1rem solid $card-border-color;
+  border-radius: 0.5rem;
```
BudnikOleksii, line 9: It's a good idea to use `rem` but in this case it doesn't work. And for border rems usually doesn't make sense

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,21 @@
+@include font;
+
+* {
+  margin: 0;
+  padding: 0;
```
BudnikOleksii, line 5: Don't set margins and padding for all elements, reset only for elements that you actually need, for example div and span doesn't have any of them

src/styles/utils/mixins.scss
```diff
@@ -0,0 +1,17 @@
...
+  font-weight: 400;
+  font-style: normal;
+}
+
+.roboto-medium {
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+  font-style: normal;
+}
+
```
BudnikOleksii, line 12: They are not mixins, just classes. And it doesn't make much sense as `Roboto` is a main font so you just set it to html or body tag and just change font weight when you need

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,9 @@
+$main-text-color: #060b35;
+$secondary-text-color: #616070;
+$button-text-color: #fff;
+$button-hover-text-color: #00acdc;
+$button-background-color: #00acdc;
+$button-hover-color: #fff;
+$card-border-color: #f3f3f3;
+$button-hover-border-color: #00acdc;
```
BudnikOleksii, line 8: You repeat `#00acdc` and white color 3 times, it should be only one, for example, `main-color`/`secondary-color` or something similar. Usually, you have a pallet and use the corresponding colors for different elements

BudnikOleksii (changes requested):
Good work 👍 
It's a good practice to use `rem`. I left some comments to check them, also something wrong with the styles in the demo, and tests failed
image

BudnikOleksii (approved):
Looks good 🔥

PR #4698, opened 2024-08-17
added task solution

mklecki:
- [DEMO LINK](https://mklecki.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://mklecki.github.io/layout_product-cards/report/html_report/)

- [x] there are 2 BEM blocks `card` and `stars` each in it own file

NOTE: also created reset.scss (shown in diff as renamed index.scss) in order to avoid using @import in main.scss or putting css for body at the end of generated css file


mklecki:
interesting test failure... well, npm test still works for me locally

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  &__buy-link {
+    margin-top: 16px;
+
+    box-sizing: border-box;
+    border: 1px solid #{$blue-accent-color};
+    border-radius: 5px;
+
+    width: 166px;
```
Zibi95, line 88: Use percentage values (e.g., 100%) for the width to ensure it spans the entire width of the container. If you change the width of the card, you will need to update this value as well if it's set to a fixed pixel value.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,107 @@
+@use '../utils/variables' as *;
+
+.card {
+  box-sizing: border-box;
+  border: 1px solid #{$gray-elements-color};
+  border-radius: 5px;
```
Zibi95, line 6: This border-radius value appears in several places with the same value, making it a good candidate for a variable.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,107 @@
...
+
+  width: 200px;
+  height: 408px;
```
Zibi95, line 9: Don't apply a fixed width or height to blocks. In this case, it's explicitly stated that the card has a width of 200px, so that's acceptable, but let the content determine the height of the card.

Zibi95 (approved):
Good job. I have left some comments for future reference. 🥇

PR #4703, opened 2024-08-19
solution

AlenaLoik:
- [DEMO LINK](https://AlenaLoik.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://AlenaLoik.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,11 +8,61 @@
     />
     <title>Product cards</title>
     <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
+      rel="stylesheet"
+    />
+    <link
       rel="stylesheet"
       href="./styles/index.scss"
     />
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div
+      class="item"
+      data-qa="card"
+    >
+      <header class="item__header">
+        <img
+          src="./images/imac.jpeg"
+          alt="item"
+        />
+      </header>
+      <main class="item__main">
+        <div class="item__title">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </div>
+        <div class="item__code">Product code: 195434</div>
+        <div class="item__reviews">
+          <div class="item__reviews_rating stars">
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+          </div>
+          <div class="item__reviews_count">Reviews: 5</div>
+        </div>
+        <div class="item__price">
+          <div class="item__price_text">Price:</div>
+          <div class="item__price_number">$2,199</div>
+        </div>
+      </main>
+      <footer class="item__footer">
+        <button
+          class="item__action"
+          data-qa="hover"
+        >
+          buy
+        </button>
+      </footer>
+    </div>
   </body>
 </html>
--- a/src/styles/blocks/item.scss
+++ b/src/styles/blocks/item.scss
@@ -1,0 +1,89 @@
+.item {
+  box-sizing: border-box;
+  width: 200px;
+  padding: 32px 16px;
+
+  border: 1px solid $border-color;
+  border-radius: 5px;
+
+  font-family: ' Roboto ', sans-serif;
+  font-size: 10px;
+  line-height: 14px;
+  font-weight: 400;
+  color: $main-text-color;
+
+  &__header {
+    box-sizing: border-box;
+    margin-bottom: 36px;
+
+    & img {
+      width: 100%;
+      height: 134px;
+      object-fit: contain;
+    }
+  }
+
+  &__main {
+    box-sizing: border-box;
+    margin-bottom: 16px;
+  }
+
+  &__title {
+    font-weight: 500;
+    font-size: 12px;
+    line-height: 18px;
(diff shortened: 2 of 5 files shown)


src/index.html
```diff
@@ -7,12 +7,62 @@
...
       href="./styles/index.scss"
     />
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div
+      class="item"
+      data-qa="card"
+    >
+      <header class="item__header">
```
TarasHoliuk, line 32: it's not a header

header tag is used for the page header
if you need wrapper you can use div, but investigate - maybe you can get rid of the wrapper, add class to the img tag, and add styles using this class

src/index.html
```diff
@@ -7,12 +7,62 @@
...
+    />
     <link
       rel="stylesheet"
       href="./styles/index.scss"
     />
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div
+      class="item"
```
TarasHoliuk, line 29:
```suggestion
      class="card"
```
it describes it better, also rename scss file accordingly

src/index.html
```diff
@@ -7,12 +7,62 @@
...
+      data-qa="card"
+    >
+      <header class="item__header">
+        <img
+          src="./images/imac.jpeg"
+          alt="item"
+        />
+      </header>
+      <main class="item__main">
+        <div class="item__title">
```
TarasHoliuk, line 39:
```suggestion
        <h2 class="item__title">
```
use semantic tag

`h1` usually is used for the page title, so let's use `h2` here

src/index.html
```diff
@@ -7,12 +7,62 @@
...
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </div>
+        <div class="item__code">Product code: 195434</div>
+        <div class="item__reviews">
```
TarasHoliuk, line 43:
```suggestion
        </h2>

        <p class="item__code">Product code: 195434</p>

        <div class="item__reviews">
```
 p tag for text
 
 add blanc line between sibling elements to increase readability

src/index.html
```diff
@@ -7,12 +7,62 @@
...
+            <div class="stars__star"></div>
+          </div>
+          <div class="item__reviews_count">Reviews: 5</div>
+        </div>
+        <div class="item__price">
+          <div class="item__price_text">Price:</div>
+          <div class="item__price_number">$2,199</div>
+        </div>
+      </main>
+      <footer class="item__footer">
```
TarasHoliuk, line 58:
```suggestion
```
wrong usage of the footer tag (same as header)
but wrapper is redundant here - just add styles you need to the button

src/styles/blocks/item.scss
```diff
@@ -0,0 +1,89 @@
+.item {
+  box-sizing: border-box;
```
TarasHoliuk, line 2: consider define it in * selector and remove everywhere else

src/styles/blocks/item.scss
```diff
@@ -0,0 +1,89 @@
...
+  width: 200px;
+  padding: 32px 16px;
+
+  border: 1px solid $border-color;
+  border-radius: 5px;
+
+  font-family: ' Roboto ', sans-serif;
```
TarasHoliuk, line 9:
```suggestion
  font-family: 'Roboto', sans-serif;
```

src/styles/blocks/item.scss
```diff
@@ -0,0 +1,89 @@
...
+  font-size: 10px;
+  line-height: 14px;
+  font-weight: 400;
```
TarasHoliuk, line 12: it's default value - no need to set it

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,19 @@
...
+
+  &__star {
+    width: 16px;
+    height: 16px;
+    box-sizing: border-box;
+    background-image: url(../images/star.svg);
+    background-repeat: no-repeat;
+    background-position: left;
+
+    &:nth-child(-n + 4) {
```
TarasHoliuk, line 15: keep styels for all possible stars count (reuse code from the "stars" task)

TarasHoliuk (changes requested):
GJ 👍 
let's improve your code a bit

src/styles/blocks/item.scss
AlenaLoik: image

src/styles/blocks/item.scss
AlenaLoik: It`s not default

etojeDenys (changes requested):
you need to pass all tests, don't hesitate to ask for some help in the chat

AlenaLoik:
> you need to pass all tests, don't hesitate to ask for some help in the chat

So sorry, I forgot make a build and deploy with updates

src/styles/blocks/item.scss
TarasHoliuk: Then you could try without quotes at all ar wit double quotes: `"Roboto"`

src/styles/blocks/item.scss
TarasHoliuk: It's just a fact) Some font-weight has to be default: [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight)

image

src/index.html
```diff
@@ -7,12 +7,59 @@
...
+        src="./images/imac.jpeg"
+        alt="card-image"
+      />
+      <main class="card__main">
+        <h2 class="card__title">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </h2>
+        <div class="card__code">Product code: 195434</div>
+        <div class="card__reviews">
+          <div class="card__reviews_rating stars stars--4">
```
TarasHoliuk, line 43:
```suggestion
          <div class="card__stars stars stars--4">
```
you could use BEM-mix - name card element same as the block - i.e. make it an element (of card block) and the block (stars) at the same time

TarasHoliuk (approved):
Nicely done! Approved ✔️ 
Added a few comments FYI

PR #4730, opened 2024-09-01
finishing card challenge

italomagno:
- [DEMO LINK](https://italomagno.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://italomagno.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,6 +6,11 @@
       name="viewport"
       content="width=device-width, initial-scale=1.0"
     />
+
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
+      rel="stylesheet"
+    />
     <title>Product cards</title>
     <link
       rel="stylesheet"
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -6,13 +6,51 @@
...
+      <div class="card__text">
+        <h2 class="card__title">
+          APPLE A1419 iMac 27" Retina
+          <br />
+          5K Monoblock (MNED2UA/A)
+        </h2>
+        <p class="card__description">Product code: 195434</p>
+      </div>
+      <div class="card__review">
+        <div class="card__stars">
```
pedro-ruas, line 35: Here you should have used the same Block from the *Stars* task.
The one with modifiers for how many stars should be active or not.

pedro-ruas (changes requested):
Italo, everything is passing the tests and your code is good, but you have missed the point in one topic from the Checklist.
The implementation for the stars rating should use the same Block/logic from the Stars task, so we can change it using modifiers.
I believe that fixing that would be simple.

src/index.html
italomagno: Hey pedro-ruas, thank you for your feedback. I had been fixing this to implement in next activity as well. Now i think that everything is fine.

pedro-ruas (approved):
Flawless!!

PR #4779, opened 2024-09-23
all tests is complete

AntonNimachuk:
- [DEMO LINK](https://antonnimachuk.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://antonnimachuk.github.io/layout_product-cards/report/html_report/)


TarasHoliuk:
Fix style lint errors:

image

src/index.html
```diff
@@ -6,13 +6,58 @@
...
+    >
+      <img
+        class="card__image"
+        src="../src/images/imac.jpeg"
+        alt="iMac 27 Retina"
+      />
+
+      <h2 class="card__title">
+        APPLE A1419 iMac 27" Retina
+        <br />
```
TarasHoliuk, line 32: There is no need to manually break title
It will automatically break into 2 lines
And with the correct font size and card width it will be done exactly the same as in test reference (and Figma)

src/index.html
```diff
@@ -6,13 +6,58 @@
...
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+        </div>
+        <p class="card__reviews-text">Reviews: 5</p>
+      </div>
+
+      <div class="card__price-block">
```
TarasHoliuk, line 49: also use `p` tag for text

src/styles/card.scss
```diff
@@ -0,0 +1,113 @@
+@import './variables';
+@import './stars';
```
TarasHoliuk, line 2: There is no need to import it here - only in the index.scss file is enough

src/styles/card.scss
```diff
@@ -0,0 +1,113 @@
...
+
+* {
+  margin: 0;
+  padding: 0;
```
TarasHoliuk, line 6: margin and padding are better to reset for the list of tags you really need (* selector is not the best choice here)

And it shouldn't exist in card.scss file - move it to a separate file or at least into index.scss (same for body styles below)

src/styles/card.scss
```diff
@@ -0,0 +1,113 @@
...
+  box-sizing: border-box;
+}
+
+body {
+  font-family: $font-family;
+  background-color: $white;
+}
+
+.card {
+  box-sizing: border-box;
```
TarasHoliuk, line 16:
```suggestion
```
already defined in `*` selector

src/styles/card.scss
```diff
@@ -0,0 +1,113 @@
...
+
+  &__image {
+    width: 160px;
+    height: 134px;
+    display: block;
+    margin: 0 auto 40px;
+  }
+
+  &__title {
+    font-family: Roboto;
```
TarasHoliuk, line 31: it will be inherited, so there is no need to define it here

But remember that you always should define fallback font-families

src/styles/card.scss
```diff
@@ -0,0 +1,113 @@
...
+
+  &__reviews {
+    color: #060B35;
+    display: flex;
+    justify-content: space-between;
+  }
+  &__reviews-text {
+    margin: 0;
+    font-size: 10px;
+    font-weight: 400;
```
TarasHoliuk, line 56: it's default value - no need to set it

src/styles/card.scss
```diff
@@ -0,0 +1,113 @@
...
+    color: #060b35;
+    align-items: center;
+  }
+
+  &__price-block {
+    margin-top: 24px;
```
TarasHoliuk, line 62: It's better to use only margin-bottom or only margin-top

src/styles/card.scss
```diff
@@ -0,0 +1,113 @@
...
+    justify-content: space-between;
+  }
+
+  &__price {
+    font-family: Roboto;
+    font-size: 12px;
+    font-weight: 400;
+    line-height: 18px;
+    text-align: left;
+    color: #616070;
```
TarasHoliuk, line 74: use variables for repeatable colors

src/styles/card.scss
```diff
@@ -0,0 +1,113 @@
...
+    align-items: center;
+    justify-content: space-between;
+  }
+
+  &__price {
+    font-family: Roboto;
+    font-size: 12px;
+    font-weight: 400;
+    line-height: 18px;
+    text-align: left;
```
TarasHoliuk, line 73: looks redundant -  justify-content: space-between; will do all work here

src/styles/card.scss
```diff
@@ -0,0 +1,113 @@
...
+      font-family: Roboto;
+      font-size: 16px;
+      font-weight: 700;
+      line-height: 18px;
+      text-align: right;
+    }
+  }
+
+  &__buy {
+    width:100%;
```
TarasHoliuk, line 87:
```suggestion
    width: 100%;
```

src/styles/card.scss
```diff
@@ -0,0 +1,113 @@
...
+    display: flex;
+    text-align: center;
+    padding: 10px 0;
+    background-color: $blue-accent;
+    color: $white;
+    text-decoration: none;
+    border-radius: 4px;
+    margin-top:16;
```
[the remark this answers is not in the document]
TarasHoliuk: good comment, fix it:
image

src/styles/stars.scss
```diff
@@ -0,0 +1,21 @@
+.stars {
+  display: flex;
+  margin: 0;
+  &__star {
```
TarasHoliuk, line 4:
```suggestion
  margin: 0;

  &__star {
```
fix all similar cases - add blank line before each selector

TarasHoliuk (changes requested):

PR #4817, opened 2024-10-05
solution files added

MaxOnRem:
- [DEMO LINK](https://MaxOnRem.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://MaxOnRem.github.io/layout_product-cards/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,11 +8,70 @@
     />
     <title>Product cards</title>
     <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
       rel="stylesheet"
-      href="./styles/index.scss"
+    />
+    <link
+      rel="stylesheet"
+      href="styles/index.scss"
     />
   </head>
   <body>
-    <h1>Product cards</h1>
+    <article
+      class="product-card"
+      data-qa="card"
+    >
+      <img
+        class="product-card__image"
+        src="images/imac.jpeg"
+        alt="Product logo"
+      />
+
+      <div class="product-card__name">
+        APPLE A1419 iMac 27" Retina
+        <br />
+        5K Monoblock (MNED2UA/A)
+      </div>
+
+      <div class="product-card__code">Product code: 195434</div>
+
+      <div class="product-card__info">
+        <div class="stars stars--4">
+          <div class="stars__star"></div>
+
+          <div class="stars__star"></div>
+
+          <div class="stars__star"></div>
+
+          <div class="stars__star"></div>
+
+          <div class="stars__star"></div>
+        </div>
+
+        <p class="product-card__review">Reviews: 5</p>
+      </div>
+
+      <div class="product-card__price">
+        <p class="product-card__price-title">Price:</p>
+
+        <p class="product-card__price-value">$2,199</p>
+      </div>
+
+      <a
+        href="#buy"
+        class="product-card__buy"
+        data-qa="hover"
+      >
+        Buy
+      </a>
+    </article>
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -7,12 +7,69 @@
...
+    >
+      <img
+        class="product-card__image"
+        src="images/imac.jpeg"
+        alt="Product logo"
+      />
+
+      <div class="product-card__name">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </div>
```
anastasiiavorobiova, line 40:
```suggestion
      <h3 class="product-card__name">
        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
      </h3>
```
Consider leveraging semantic tags for titles

src/index.html
```diff
@@ -7,12 +7,69 @@
...
+
+      <div class="product-card__code">Product code: 195434</div>
```
anastasiiavorobiova, line 42:
```suggestion
      <p class="product-card__code">Product code: 195434</p>
```
Consider leveraging semantic tag for text

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,20 @@
...
+    height: 16px;
+    background-repeat: no-repeat;
+    background-position: center;
+    background-image: url(../images/star.svg);
+  }
+  &__star:not(:last-child) {
+    margin-right: 4px;
+  }
+
+  &--4 {
```
anastasiiavorobiova, line 15: Consider adding styles for all possible modifiers

anastasiiavorobiova (changes requested):
Great start! Make sure to deploy your solution

src/styles/index.scss
```diff
@@ -1,3 +1,14 @@
+@import './utils/variables';
+@import './blocks/stars';
+@import './blocks/card';
+
 body {
+  font-family: Roboto, sans-serif;
+  font-weight: 400;
+  line-height: 18px;
```
volodymyr-soltys97, line 8: Move font styles to the html selector

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
This part of the task is incomplete, you need to use styles from the previous task 
image
image

src/styles/index.scss
MaxOnRem: Sorry but I I can't understand what these styles are in your screen, project deployment provides styles from the previous task, with first 4 active stars.
Agree with styles should be moved to html selector, and even AI-mentor told me this, but could you show the rule in our code styles rules that I can be sure further

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,41 @@
...
+
+  &--1 {
+    .stars__star:nth-child(-n + 1) {
+      background-image: url(../images/star-active.svg);
+    }
+  }
+  &--2 {
+    .stars__star:nth-child(-n + 2) {
+      background-image: url(../images/star-active.svg);
+    }
```
volodymyr-soltys97, line 24: Fix it everywhere
```suggestion
  &--1 &__star:nth-child(-n + 1),
  &--2 &__star:nth-child(-n + 2),
  ........ {
      background-image: url(../images/star-active.svg);
    }
  }
```

volodymyr-soltys97 (changes requested):
Hi, the comment from the previous review still not fixed, you need to implement the modifier logic

src/styles/index.scss
TarasHoliuk: Not sure such a rule exists in `code styles rules` it's not about code style (at least not only). I guess it was mentioned in the theory and video about fonts. And it's a general rule - font-family (and similar to font-weight) should be the same for the entire page (with some exceptions, in these cases it overrides in places it is needed) and it will be inherited. So there is no need to write it for each tag with text

TarasHoliuk (approved):

PR #4840, opened 2024-10-10
add solution

dHyba:

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,6 +6,18 @@
       name="viewport"
       content="width=device-width, initial-scale=1.0"
     />
+    <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
+      rel="stylesheet"
+    />
     <title>Product cards</title>
     <link
       rel="stylesheet"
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -6,13 +6,71 @@
...
   <body>
-    <h1>Product cards</h1>
+    <div
+      class="card"
+      data-qa="card"
+    >
+      <img
+        class="card__img"
+        src="./images/imac.jpeg"
+        alt="Product"
```
BudnikOleksii, line 35: alt attribute should describe the image content
image

src/index.html
```diff
@@ -6,13 +6,71 @@
...
+
+      <section class="card__info">
+        <div class="card__title">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </div>
+
+        <span class="card__code">Product code: 195434</span>
+
+        <div class="card__reviews">
+          <div class="stars stars--rating-4">
```
BudnikOleksii, line 46: Use semantic tags where possible

src/index.html
```diff
@@ -6,13 +6,71 @@
...
+            <div class="stars__star stars__star--active"></div>
+
+            <div class="stars__star stars__star--active"></div>
+
+            <div class="stars__star stars__star--active"></div>
+
+            <div class="stars__star stars__star--active"></div>
```
BudnikOleksii, line 53: You don't need active class, you have to set active stars depending on the block modifier. If the modifier is `stars--3` you should show 3 active start, if `stars--4` - 4 and so on

src/styles/blocks/star.scss
```diff
@@ -0,0 +1,16 @@
...
+    background-image: url('../images/star.svg');
+    background-repeat: no-repeat;
+    background-position: center;
+    margin-right: 4px;
+  }
+}
+
+.stars--rating-4 .stars__star:nth-child(-n + 4) {
+  background-image: url('../images/star-active.svg');
+}
```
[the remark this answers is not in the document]
BudnikOleksii: > It would be a good idea to provide styles for all possible star ratings (from 1 to 5). This way, you can easily change the rating by just modifying the class of the parent element. For example, you can add styles for `.stars--rating-1`, `.stars--rating-2`, etc., each with the appropriate number of active stars.

Good point

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions, check them. Also, please, add a demo link to the PR description so a reviewer can test your solution

PR #4959, opened 2024-11-05
did the mentor task and made a branch develop

IlyaPerfomance:


SanyaBratashchuk (changes requested):
there should be just 2 BEM blocks card and stars others are card's elements

PR #5069, opened 2024-12-15
add task solution

Maks-Chervyakov:


src/index.html
```diff
@@ -7,12 +7,60 @@
...
       href="./styles/index.scss"
     />
   </head>
   <body>
-    <h1>Product cards</h1>
+    <div
+      class="card"
+      data-qa="card"
+    >
+      <div class="card__image"></div>
```
vadiimvooo, line 33: Use img tag to display card image instead of doing it with background property

vadiimvooo (changes requested):
Good job, a few changes to fix.

src/index.html
Maks-Chervyakov: fixed

VitaliyBondarenko1982 (approved):
Looks good.
Do not forget add demo links to PR description in next tasks.

PR #5077, opened 2024-12-17
add task solution

wiltonmartinsdev:
- [DEMO LINK](https://wiltonmartinsdev.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://wiltonmartinsdev.github.io/layout_product-cards/report/html_report/)


src/index.html
```diff
@@ -6,13 +6,86 @@
...
+          class="product-card__image"
+        />
+      </div>
+      <h1 class="product-card__title">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </h1>
+      <h2 class="product-card__code">Product code: 195434</h2>
+
+      <div class="product-card__reviews">
+        <svg
```
pedro-ruas, line 46: What is this svg doing here?
You should have used the same code as for the Stars task done before. The stars should be added using css and work exactly as the previous task (you can literally copy-paste most of it)

src/styles/index.scss
```diff
@@ -1,3 +1,124 @@
...
+    margin-bottom: 30px;
+  }
+
+  &__title {
+    font-size: 12px;
+    font-weight: 500;
+    line-height: 18px;
+    text-align: left;
+
+    width: 166px;
```
pedro-ruas, line 44: We avoid using these broken numbers. If you saw a number like this in the design, it must certainly comes from another styling, like `width: 100%` or something like that

src/styles/index.scss
```diff
@@ -1,3 +1,124 @@
...
+  }
+
+  &__price {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+  }
+
+  &__star {
+    width: 20px;
```
pedro-ruas, line 69: I don't believe this is the right size.
And as I said before, you can use the code from Stars' task

src/styles/index.scss
```diff
@@ -1,3 +1,124 @@
...
 body {
   margin: 0;
+
+  color: $main-accent-color;
+
+  font-family: Roboto, sans-serif;
+  font-size: 16px;
+}
+
+.product-card {
```
pedro-ruas, line 14: I believe the height of your card is wrong and it is causing most fails in the test

pedro-ruas (changes requested):
There are some fixes needed, and the styling is not passing the auto tests

src/styles/index.scss
```diff
@@ -1,3 +1,14 @@
+@import 'variables';
+@import 'product-card';
+@import 'stars';
+
```
joaorpereira, line 4: Missing declare: 

```
* {
    margin: 0;
    padding: 0;
    box-sizing: border-box;
}
```

https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing

joaorpereira (changes requested):
On the left is what the tests expect to see, and on the right is the your work

image

src/styles/index.scss
wiltonmartinsdev: Thank you, i will make the adjustments.

src/styles/index.scss
wiltonmartinsdev: Thanks for the tip, I'll make the adjustments

pedro-ruas (approved):
Now we are talking!

PR #5085, opened 2024-12-18
Develop

pat5513:
- [DEMO LINK](https://pat5513.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://pat5513.github.io/layout_product-cards/report/html_report/)

- [X] there 2 BEM blocks `card` and `stars` each in their own file
- [X] SCSS Nesting is used for `elements`, `modifiers` and `pseudo-classes`
- [X] SCSS Variables are used for main values and placed in a **separate file**
- [X] all `stars--N` modifiers work as expected (Highlight first `N` stars)
- [X] Code follows all the [Code Style Rules ❗️](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules)

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,5 @@
+body {
+  margin: 0;
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+}
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,9 +1,11 @@
+@use '../utils/variables' as *;
+
 .card {
   box-sizing: border-box;
   padding: 16px;
   width: 200px;
-  border: 1px solid black;
-  border-radius: 10px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
   display: flex;
   flex-direction: column;
   justify-content: center;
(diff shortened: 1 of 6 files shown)


natalia-klonowska (changes requested):
to pass tests you need to:
- change font-weight for text "Price:" to 400
- all colors for texts are not black but #060B35
- margin between card__code and card__stars is 14px instead of 16px because of margin collapse (https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_box_model/Mastering_margin_collapsing) best way to fix it is by using only top or only bottom margins for every element inside card

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,80 @@
...
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+
+  &__image {
+    width: 160px;
+    height: 134px;
+    margin: 16px 3px;
```
natalia-klonowska, line 16: fix margin collapse and center image without adding margins to the sides

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,80 @@
...
+
+  &__total {
+    font-size: 1.6em;
+    font-weight: $main-weight;
+    line-height: $line;
+  }
+
+  &__button {
+    height: 40px;
+    width: 166px;
```
natalia-klonowska, line 62: use 100% so your code would be more flexible and responsive ;)
```suggestion
    width: 100%;
```

natalia-klonowska (changes requested):
Great work 🚀 The only margin collapse left is on the image element

natalia-klonowska (approved):

PR #5253, opened 2025-01-30
 product card task

Iwantschenko:
❗️ Replace Iwantschenko with your GitHub username and copy the links to the Pull Request description:

[DEMO LINK](https://iwantschenko.github.io/layout_product-cards/)
[TEST REPORT LINK](https://iwantschenko.github.io/layout_product-cards/report/html_report/)
❗️ Copy this Checklist to the Pull Request description after links, and put - [x] before each point after you checked it.

 there 2 BEM blocks card and stars each in their own file
 SCSS Nesting is used for elements, modifiers and pseudo-classes
 SCSS Variables are used for main values and placed in a separate file
 all stars--N modifiers work as expected (Highlight first N stars)
 Code follows all the [Code Style Rules ❗️](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules)

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,85 @@
+.card {
+  display: flex;
+  flex-direction: column;
+
+  box-sizing: border-box;
+  width: $card-width;
+  padding: 16px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+
+  &__img {
+    height: 134px;
+    width: 160px;
+    margin: 16px auto 40px;
+    object-fit: cover;
+  }
+
+  &__content {
+    font-size: 10px;
+    font-weight: 400;
+    line-height: 14px;
+    color: $card-main-color;
+  }
+
+  &__description {
+    margin-bottom: 4px;
+    font-size: 12px;
+    font-weight: 500;
+    line-height: 18px;
+  }
+
+  &__code {
+    color: $card-secondary-color;
+    margin-bottom: 16px;
+  }
+
+  &__rating {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    margin-bottom: 24px;
+  }
+
+  &__price {
+    display: flex;
+    justify-content: space-between;
+    line-height: 18px;
+    margin-bottom: 16px;
+  }
+
+  &__button {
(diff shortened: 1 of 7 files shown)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,86 @@
...
+    text-align: center;
+    color: vars.$card-button-hover-color;
+    background-color: vars.$card-button-color;
+    border-radius: 5px;
+  }
+
+  &__link:hover {
+    color: vars.$card-button-color;
+    background-color: vars.$card-button-hover-color;
+    border: 1px solid vars.$card-button-color;
```
alexander-ignatow, line 74: This is the problem - you add the border on hover, which makes the button "jump"

alexander-ignatow (changes requested):
Good job, but please fix the issues below (the button changes size on hover):

image

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,85 @@
...
+    background-color: $card-button-color;
+    border-radius: 5px;
+  }
+
+  &-link:hover {
+    color: $card-button-color;
+    box-sizing: border-box;
+    background-color: $card-button-hover-color;
+    border: 1px solid $card-button-color;
+  }
```
alexander-ignatow, line 84: The solution is not optimal - I would suggest using the following: 

```suggestion
  &-link {
    line-height: $card-button-height;
    text-decoration: none;
    font-size: 14px;
    font-weight: 700;
    text-transform: uppercase;
    text-align: center;
    color: $card-button-hover-color;
    background-color: $card-button-color;
    border-radius: 5px;
    border: 1px solid $card-button-color;
  }

  &-link:hover {
    color: $card-button-color;
    background-color: $card-button-hover-color;
  }
```

0. Make sure that button height is correct. Don't use height. You may keep line-height, use font size + padding - that is the recommended way.

1. Set the border for the link and never change it. 
2. Changing only background and font color does all the job.
3. Avoid changing box-sizing on hover. I would suggest just avoid using it at all.
4. General rule - change as little styles as possible.

alexander-ignatow (approved):
Lets move on, but please, mind my comments on the button 👇

PR #5637, opened 2025-05-29
feat: ajuste no tamanho da imagem do card do produto

FelipeLeal2021:


src/index.html
```diff
@@ -1,18 +1,31 @@
...
+      <img src="images/imac.jpeg" alt="iMac" />
+    </section>
+
+    <h1 data-qa="product-title">
+      APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+    </h1>
+
+    <section class="product-info">
+      <div class="stars stars--4" data-qa="product-rating">
+        <img src="images/star-active.svg" alt="star-active" />
```
pedro-ruas, line 20: You should use the same block from your previous task, Stars, this is not it

pedro-ruas (changes requested):
You should use the command `npm run deploy` and add your links, both TEST and DEMO to your pull request.
I can see the tests failed, but I also need to check the page.
There is also an issue in the way you implemented the stars part

PR #5899, opened 2025-08-18
Solution

Pavlo-Dv:


maxim2310 (approved):
Good job, but plz attach test link also next time

PR #6007, opened 2025-09-16
as

VitaliyHoroshko:


etojeDenys (changes requested):
it seems that you are not deploying your solution. feel free to ask for help in the chat if you have any problems

etojeDenys (changes requested):
demo link looks unfinished, and the test link leads to an error page. please fix these issues and then request the review again
image

PR #6013, opened 2025-09-17
Develop Layour_cards

anastasiia-levochkina:
[DEMO LINK](https://anastasiia-levochkina.github.io/layout_product-cards/)
[TEST REPORT LINK](https://anastasiia-levochkina.github.io/layout_product-cards/report/html_report/)


Moroz-Dmytro (changes requested):
Please fix all coments from AI
And also check tests, there is difference between target and your solution

PR #6255, opened 2025-12-19
add solution

thmoro:
- [x] there 2 BEM blocks `card` and `stars` each in their own file


thmoro:
@mateacademy-ai-mentor about the background-image paths inside the star block, I can't change, because it needs to take index.scss as the css root, and they are not at the same level.
Now I'll push the other modifications mentioned.

PR #6353, opened 2026-02-09
Product cards layout with BEM and SCSS

Wallace775:


pedro-ruas (changes requested):
There is no code in this PR

PR #6541, opened 2026-04-28
Implement product card layout

Igorls34:


IanLuan (changes requested):
Hi Igor, your code looks fine but it's missing the DEMO link. Please run the deploy.

src/styles/blocks/_card.scss
```diff
@@ -0,0 +1,90 @@
+@import '../utils/variables';
+
+.card {
+  box-sizing: border-box;
+  width: $card-width;
+  padding: 32px 16px 16px;
```
raulriato, line 6: You don't need to use `16px` 2 times. If you use `32px 16px` it already uses 32px vertically and 16px horizontally.

src/styles/blocks/_card.scss
```diff
@@ -0,0 +1,90 @@
...
+    height: 134px;
+    margin: 0 auto 40px;
+    object-fit: contain;
+  }
+
+  &__title {
+    margin: 0 0 4px;
+    font-size: 12px;
+    font-weight: 500;
+    line-height: 18px;
```
raulriato, line 24: You should not copy all styles from the `mockup`. some styles, like `letter-spacing`, `line-height` are not necessary.

raulriato (approved):
You have some minor issues in your code, but nothing major, so I'll approve it.

PR #6582, opened 2026-05-20
Develop

SylwiaLeung:
- [DEMO LINK](https://SylwiaLeung.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://SylwiaLeung.github.io/layout_product-cards/report/html_report/)

- [x] there 2 BEM blocks `card` and `stars` each in their own file


natalia-klonowska (changes requested):
- redeploy your project (npm run deploy command) because styles didn’t load properly

Image

- add border 
Image

src/styles/card.scss
```diff
@@ -0,0 +1,83 @@
...
+.card {
+  padding: 207px 17px 17px;
+  box-sizing: border-box;
+  width: 200px;
+  background: {
+    image: url(../images/imac.jpeg);
+    size: 160px 134px;
+    repeat: no-repeat;
+    position: center 33px;
+  }
```
natalia-klonowska: It’s better to use img instead of background-image for product photo because img improves accessibility, SEO and allows better support for features like lazy loading and responsive images. background-image is intended for purely decorative visuals and is ignored by screen readers. 
[more info](https://medium.com/@aidana1529/what-should-you-choose-img-or-background-image-f9bebc99180)

src/styles/stars.scss
```diff
@@ -0,0 +1,23 @@
...
+  display: flex;
+
+  &__star {
+    padding: 8px;
+    background: {
+      image: url(../images/star.svg);
+      repeat: no-repeat;
+      position: center;
+      size: 82%;
+    }
```
natalia-klonowska, line 12: how did you come up with this padding and size? It would be simpler and clearer to set width and height directly from design spec (16px)

src/utils/_variables.scss
```diff
@@ -0,0 +1,15 @@
...
+$font-sm: 10px;
+$font-md: 12px;
+$font-lg: 14px;
+$font-default: 16px;
+$line-height-sm: 14px;
+$line-height-md: 16px;
+$line-height-lg: 18px;
+$vertical-margin-sm: 4px;
+$vertical-margin-md: 16px;
+$vertical-margin-lg: 24px;
```
natalia-klonowska: Variables should be used for values that are repeated and connected by purpose. there’s no need to create variables for every padding or font size. that can actually reduce readability because instead of seeing value directly you have to keep jumping between different files.

natalia-klonowska (changes requested):

src/styles/card.scss
KanapowyZiemniak: The instructions include the sentence 

> - `background-image: url()` should be relative to the `index.scss`. So it should start with `../images`.

so I inferred that it must be referring to this image — that it’s supposed to be in the background. But if it’s not referring to this image, then I don’t understand what the instruction is.

src/styles/stars.scss
KanapowyZiemniak: Actually, it didn’t occur to me that I could just set the width/height instead of handling it with padding. Same with size - I didn’t realise I could simply enter the size value. Thanks!

src/utils/_variables.scss
KanapowyZiemniak: Cool, thanks

src/styles/card.scss
natalia-klonowska: That instruction referred to stars that you add as background-image

natalia-klonowska (approved):
Great work! 🚀 just remember to update demo (npm run deploy) every time you add changes 

Image

PR #6631, opened 2026-06-15
Update README with task breakdown

luis-botelho:
Adds a folder-structure guide and per-file task notes in the README, without changing the exercise implementation.

Note: there is still an unrelated local change in `src/index.html` that was not included in this commit.


DanielCamposDevX (changes requested):
This task is wrongly assigned you should use your develop branch. Missing deploy link and test report link

PR #6642, opened 2026-06-21
Add product card layout and styles with responsive design

luis-botelho:


DanielCamposDevX (changes requested):
Hi luis! You are missing some key features ok? 
You need to remember to correct your tests, run npm run test so you can know what it needs to be changed.

After correcting the tests make npm run deploy, after that edit the description with Deploy link and Demo link.


The corrections above follow the https://mate-academy.github.io/layout_task-guideline/ .

DanielCamposDevX (changes requested):
Still no demo link or test Report Link Luis! 
it seems like your deploy is not working, remember to verify before submitting a new review

luis-botelho:
Now is really solved

> Still no demo link or test Report Link Luis! it seems like your deploy is not working, remember to verify before submitting a new review

src/styles/_variables.scss
```diff
@@ -0,0 +1,9 @@
+$primary-color: #fff;
+$border-color: #f3f3f3;
+$secondary-color: #00acdc;
+$button-hover-color: #0a7ea8;
```
raulriato, line 4: This is wrong. The color showed in the `mockup` is different from the one you're using.

raulriato (approved):
The tests are passing, somehow, so I'll approve it.

But I really recommend you taking a look on the `hover` background color you used in the `button`, because it's not correct and you will need to change it for next projects where you use this card.

PR #6686, opened 2026-07-14
add task solution

eurobertinhosantos:
- [DEMO LINK](https://eurobertinhosantos.github.io/layout_product-cards/)
- [TEST REPORT LINK](https://eurobertinhosantos.github.io/layout_product-cards/report/html_report/)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,80 @@
+.card {
+  box-sizing: border-box;
+  width: $card-width;
+  padding: $card-padding-top $card-padding-inline $card-padding-bottom;
```
raulriato, line 4: This is an overuse of variables. You should create variables only for values that are reused. Since this value is used only once, keeping it inline makes the code simpler.

raulriato: Please check all the variables to see if each one of them makes sense in your code.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,80 @@
...
+  border: $card-border-width solid $color-card-border;
+  border-radius: $card-border-radius;
+  background-color: $color-card-background;
```
raulriato, line 7: You should use flexbox for the card and take advantage of its features, such as align-items, to keep the elements centered.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,80 @@
...
+
+  &__image {
+    display: block;
```
raulriato, line 10: If you'd used `flex` for the `card`, this would not be necessary.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,80 @@
...
+    width: $image-width;
+    height: $image-height;
+    margin: 0 auto 40px;
```
raulriato, line 13: Since you are not defining margin for `top`, `left` and `right`, it's better to just use `margin-bottom`

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  &__title {
+    margin: 0 0 4px;
```
raulriato, line 17: Same here, just use `margin-bottom`

raulriato: Check other styles to correct them for this.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,80 @@
...
+    color: $color-text-primary;
+    font-size: 12px;
+    font-weight: 500;
+    line-height: 18px;
```
raulriato, line 21: You should not just copy every style from the `mockup`. the property `line-height` is not necessary for any element

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  &__code {
+    margin-bottom: 16px;
+  }
+
+  &__review-container {
+    display: flex;
+    justify-content: space-between;
+    align-items: flex-end;
```
raulriato, line 31: This is probably causing some differences in the tests

src/index.html
```diff
@@ -7,12 +7,65 @@
...
+        alt="iMac A1419"
+      />
+
+      <h2 class="card__title">
+        APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+      </h2>
+
+      <div class="card__code">Product code: 195434</div>
+
+      <div class="card__review-container">
```
raulriato, line 45: This is not following `BEM` correctly.

The pattern should be either:

```html
<div class="card__review">
  <div class="stars stars--4">
      ...
    </div>
    <div class="card__review-count">...</div>
</div>
```

or

```html
<div class="card__review-container">
  <div class="stars stars--4">
      ...
    </div>
    <div class="card__review-container-count">...</div>
</div>
```

I recommend the first one.

src/index.html
```diff
@@ -7,12 +7,65 @@
...
+          <div class="stars__star"></div>
+        </div>
+
+        <div class="card__review-count">Reviews: 5</div>
+      </div>
+
+      <div class="card__price-container">
+        <div class="card__price-label">Price:</div>
+        <div class="card__price">$2,199</div>
+      </div>
```
raulriato, line 60: This has the same problem as the `card__review`

raulriato (changes requested):
You're almost there, but I'll need to recommend some changes, since you missed some important rules.

Good luck.

Remember to run `npm run deploy` after pushing the changes, so the `DEMO LINK` keeps up with your code. 

Also, remember not to close this `pull request`, just push the changes. The `pull request` will update automatically.

If you need any help, please join a `Bate papo com o mentor` meeting, or come to the [chat](https://mate.academy/pt-br/chats/39089) and ask for help.

raulriato (changes requested):
Your `DEMO LINK` is not working.

Make sure the links work properly before re-requesting review.

If you need any help, please join a `Bate papo com o mentor` meeting, or ask for help in the chat.

raulriato (changes requested):
You'd followed the recommendations all correctly, but then you changed you `index.html` again to return the classes to old names. What happened? I can't approve it this way because the `BEM` is not being followed correctly.

DanielCamposDevX (approved):

